PR #10, opened 2026-09-17
feat(identity): add the issuer model with its verification states

### LINKED ISSUE #3: [IDENTITY] TP-022 Define the issuer model
## Objective

Define the `issuer` table: the organisation that registers products and whose claims the passport reports.

## Scope

**In scope:**

- Drizzle schema for `issuer`
- Verification status as an explicit, closed set of states
- Migration

**Out of scope:**

- The verification workflow itself
- Authentication and API keys (TP-141)

## Context

The issuer is the root of the entire trust model. Per ADR 0003, TrustPass cannot prove that a physical product is genuine; it can only prove that a specific issuer claimed it is. That makes "who is this issuer, and what have we actually checked about them" the single most load-bearing fact in the system.

Verification status therefore cannot be a boolean. "We have their legal registration on file" and "we have verified nothing" and "we verified them and later suspended them" are different states with different consequences for what the passport is allowed to display.

## Acceptance criteria

- [ ] `issuer` table with company name, legal name, country and verification status
- [ ] Verification status is a database enum, not free text
- [ ] States cover at minimum: unverified, pending, verified, suspended
- [ ] Country stored as ISO 3166-1 alpha-2
- [ ] Timestamps stored with timezone
- [ ] Migration generated and applied cleanly against a fresh database
- [ ] Types exported for consumers
- [ ] A test proves an issuer round-trips and that an invalid status is rejected

## Dependencies

None.

## Component

db

## Risk

high — every claim the passport displays inherits its credibility from this record.

## Estimate

3

clevervi:
## What changed

Adds the `issuer` table — the organisation that registers products and whose
claims a passport reports — plus two ADRs recording decisions this table is the
first to apply.

## Why

Closes #3 (TP-022).

Per [ADR 0003](docs/adr/0003-identity-is-not-authenticity.md), TrustPass cannot
prove a physical product is genuine. It can only prove that a specific issuer
claimed it is. That makes *what we have actually checked about this issuer* the
single most load-bearing fact in the system, and the reason this is `risk:high`.

**1. Verification status is an enum, not a boolean** — [issuer.ts:19-31](packages/db/src/schema/issuer.ts#L19-L31)

"We hold their legal registration", "we hold nothing" and "we verified them and
later revoked it" are different facts with different consequences for what a
passport may display. A boolean collapses the third into one of the first two.
New issuers default to `unverified`, which is the honest default.

**2. Issuers are keyed by registration number, not by name** — [ADR 0006](docs/adr/0006-issuers-are-identified-by-registration-number.md)

This PR originally enforced uniqueness over `(legal_name, country)`, on the
assumption that registries issue one legal name per jurisdiction.

That assumption was checked. It **holds in Colombia** — the RUES runs national
*control de homonimia*, strict enough that the corporate form is not a
differentiator: `Carnes y Carnes S en C` counts as identical to
`Carnes y Carnes Ltda`. It **fails in the United States**, where names are
registered per state, so two unrelated companies may both be `Acme LLC`.

The data model is international from day one, so a constraint resting on one
country's naming law breaks on the first issuer from a country with different
law — by refusing a legitimate business, which is the worst direction to fail in.

Identity now comes from the number the issuer's own national authority
guarantees unique: NIT, RFC, EIN, VAT. It does not change when a company
rebrands, and it is what the verification workflow has to check anyway, since
verification is a registry lookup by number rather than by name.

`not null`, deliberately: Postgres permits unlimited nulls in a unique index, so
nullable would switch duplicate protection **off** for exactly the unverified
issuers where duplicates are most likely.

**3. Key strategy** — [ADR 0005](docs/adr/0005-internal-keys-are-never-public.md)

This answers the scalability question raised on #8. ADR 0004 accepted a cost:
128-bit random identifiers insert at random positions in a B-tree, which hurts at
the millions of rows this system is meant to hold. That cost only exists if the
public identifier is *also* the primary key.

One rule now applies everywhere: **every table has a `bigint identity` primary
key, and that key never crosses the API boundary.** Public identifiers are
separate columns. Inserts land at the end of the index, row counts stay
internal, and there is nothing to memorise per table.

## Acceptance criteria

- [x] Table with company name, legal name, country and verification status
- [x] Verification status is a database enum, not free text
- [x] States cover unverified, pending, verified and suspended
- [x] Country stored as ISO 3166-1 alpha-2
- [x] Migration generated and applied cleanly against a fresh database
- [x] Types exported for consumers
- [x] Tests prove round-trip and that an invalid status is rejected

Beyond the stated criteria: `registration_number`, because how an issuer is
uniquely identified *is* the issuer model, and the first answer was wrong.

## Tests

18 integration tests against real Postgres. Constraints only exist in the
database, so asserting them against a mock would test the mock. 52 pass overall.

Every constraint asserts its **SQLSTATE by value** — `23514` check, `23505`
unique, `23502` not-null, `22P02` invalid enum — so no test can pass because the
insert failed for an unrelated reason such as a dropped connection. Getting there
surfaced something worth knowing: Drizzle wraps driver failures, so the SQLSTATE
is not on the error that surfaces, it is down the `cause` chain.

Two defects were found during self-review and fixed in this branch; both now have
regression tests. See the review comments below.

Also adds the migration step CI was missing: the test job started Postgres and
never applied a schema.

## Security impact

The verification state machine is the foundation of every claim a passport
displays, so this table is where trust is rooted.

- `suspended` is a distinct state, not the absence of `verified`. Trust withdrawn
  is not trust never granted, and the schema keeps them apart.
- Country and registration-number formats are validated where they cannot be
  bypassed.
- Per ADR 0005 the identity key must never be serialised outside the system. The
  type system does not enforce that. It is a review concern, which is why API
  surface changes stay `risk:high`.

Not defended against, and it cannot be at this layer: a well-formed registration
number belonging to a real company the issuer does not own. That is
impersonation, and closing it is the verification workflow's entire job.

## Breaking changes

None. First migration in the project; applies cleanly to an empty database.

---

- [ ] CI is green
- [x] `pnpm-lock.yaml` is committed if any dependency changed — no dependency changed


clevervi:

Per the policy merged in #9, this is a second pass by the same judgement that
wrote the diff. It is posted for the record, not as an approval. `risk:high`
still needs a human.

It found one real defect, now fixed in `ee78ab1`.

### Defect — the uniqueness constraint did not do what this PR claimed

The description above says the index prevents duplicate issuers. It did not. A
plain Postgres unique index is case sensitive, so I checked against the running
database rather than reasoning about it:

```
INSERT INTO issuer (company_name, legal_name, country) VALUES ('A', 'CASE PROBE SAS', 'CO');
INSERT INTO issuer (company_name, legal_name, country) VALUES ('B', 'case probe sas', 'CO');

 id |   legal_name   | country
----+----------------+---------
 78 | CASE PROBE SAS | CO
 79 | case probe sas | CO
(2 rows)
```

Both inserted. The constraint prevented nothing anyone would actually type,
against a table whose entire purpose is being the root of the trust model.

Fixed by indexing `lower(legal_name)`. Migration 0000 is amended rather than
superseded — it has not been merged or deployed, so there is nothing to migrate
from.

A known limit is now recorded in the schema: whitespace and punctuation variants
still pass (`ANDES  TECH` with two spaces). Catching those means comparing
against a company registry, which is the verification workflow's job, not an
index's.

### Weak assertion — a test that could not fail

The `updated_at` test used `toBeGreaterThanOrEqual`, which passes even when
`$onUpdate` never fires. That is precisely what it exists to test. Now strictly
greater, with a short delay so millisecond resolution cannot mask it, and
`created_at` asserted unchanged.

### Reviewed and found sound

- The enum states are exhaustive for what the verification workflow needs, and
  `suspended` is correctly distinct from `unverified`
- The country check constraint rejects `co`, `C1` and whitespace, tested
- SQLSTATE assertions are by value, so a test cannot pass on an unrelated failure
- The migration applies cleanly to an empty database, exercised on a dropped and
  recreated schema

### Open for the human reviewer

The `legal_name` unique-per-country assumption. I believe company registries
issue one legal name per jurisdiction, but I have not verified that against
Colombia's RUES specifically. If it is wrong, a legitimate issuer cannot
register, and the remedy is dropping one index.

packages/db/src/schema/issuer.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * National business identifier: NIT in Colombia, RFC in Mexico, EIN in the
+     * United States, VAT number across the EU.
+     *
+     * This is the issuer's identity, not its name — see ADR 0006. Names are
+     * only unique where naming law makes them so, and they change when a
+     * company rebrands. A key that changes is not a key.
+     */
+    registrationNumber: varchar("registration_number", { length: 50 }).notNull(),
```
clevervi, line 53: This column replaced a unique index on `(legal_name, country)`.

I checked the assumption rather than trusting it. Colombia's RUES does enforce national name uniqueness, strictly enough that `S en C` and `Ltda` are not differentiators. But US names are registered per state — two real companies can both be `Acme LLC`. The old index would have refused the second one, and refusing a legitimate business is the worst way for a trust platform to fail.

Names also change when a company rebrands, which made a rebrand look like a new issuer and split a product's history in two.

clevervi (commented):
Went back over this with fresh eyes before handing it to you. Two things were wrong and are fixed on the branch; notes are inline where they belong.

One caveat worth stating plainly: I wrote this diff, so this is a second pass by the same judgement, not a second opinion. It is why `risk:high` still waits for you.

packages/db/src/schema/issuer.ts
clevervi, line 53: `notNull` is load-bearing, not tidiness. Postgres allows unlimited nulls in a unique index, so an optional registration number would turn duplicate protection **off** for exactly the `unverified` issuers where duplicates are most likely.

packages/db/src/schema/issuer.ts
```diff
@@ -0,0 +1,100 @@
...
+    // Looking an issuer up by name stays a normal thing to do; it is simply not
+    // how identity is established. Lower-cased so the lookup is case tolerant.
+    index("issuer_legal_name_idx").on(sql`lower(${table.legalName})`),
+
+    // Shaped like a registration number, without teaching the schema the rules
+    // of every tax authority on earth. Accepts 900123456-7, ABC123456T1A and
+    // 12-3456789. Whether the number is real is what verification_status exists
+    // to answer.
+    check(
+      "issuer_registration_number_format",
```
clevervi, line 88: Deliberately shape-only. It accepts `900123456-7`, `ABC123456T1A` and `12-3456789` without teaching the schema every tax authority's rules.

Whether the number is real — or belongs to someone else — is unanswerable here. That is what `verification_status` exists for, and why `unverified` is the default.

packages/db/src/schema/issuer.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * What TrustPass has actually checked about an issuer.
+ *
+ * Not a boolean, because "we hold their legal registration", "we hold nothing"
+ * and "we checked and later revoked it" are different facts with different
+ * consequences for what a passport is allowed to display. Per ADR 0003 the
+ * issuer is the root of the trust model: every claim a passport shows inherits
+ * its credibility from this column.
+ */
+export const issuerVerificationStatus = pgEnum("issuer_verification_status", [
```
clevervi, line 22: Four states rather than a boolean. `suspended` has to be distinct from `unverified`: trust withdrawn is not trust never granted, and collapsing them would erase the reason an issuer was suspended.

packages/db/src/schema/issuer.integration.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      const [updated] = await db
+        .update(issuer)
+        .set({ verificationStatus: "verified" })
+        .where(eq(issuer.id, created?.id as number))
+        .returning();
+
+      expect(updated?.verificationStatus).toBe("verified");
+      // Strictly greater, not "greater or equal": the weaker assertion passes
+      // even when $onUpdate never fires, which is the thing under test.
+      expect(updated?.updatedAt.getTime()).toBeGreaterThan(created?.updatedAt.getTime() as number);
```
clevervi, line 133: This was `toBeGreaterThanOrEqual`, which passes even when `$onUpdate` never fires — the exact thing it exists to test. Strictly greater now, with a delay so millisecond resolution cannot hide it.

packages/db/src/schema/issuer.integration.test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+      const [mexican] = await db
+        .insert(issuer)
+        .values(build({ registrationNumber: shared, country: "MX" }))
+        .returning();
+
+      expect(mexican?.country).toBe("MX");
+    });
+
+    it("allows two issuers to share a legal name in one country", async () => {
```
clevervi, line 172: The regression test for the US case. Without this, nothing would have caught the constraint being wrong until a real American issuer tried to register.

PR #24, opened 2026-09-17
feat(identity): add the product model

### LINKED ISSUE #4: [IDENTITY] TP-021 Define the product model
## Objective

Define the `product` table: the record a TrustPass ID resolves to.

## Scope

**In scope:**

- Drizzle schema for `product`
- Migration
- Inferred insert and select types exported from `@trustpass/db`

**Out of scope:**

- Registration endpoint (TP-023)
- Duplicate serial enforcement (TP-024)
- Status transitions (TP-025)

## Context

Initial fields: TrustPass ID, brand, model, serial, category, status, issuer, created at.

The serial is the manufacturer's identifier and is not ours to change. It is also sensitive: a full serial exposed publicly lets a third party impersonate the product elsewhere, or file a warranty claim against it. Storage and public display are separate concerns and the schema should not assume they are the same.

## Acceptance criteria

- [ ] `product` table defined in Drizzle with the fields above
- [ ] TrustPass ID is the primary key
- [ ] Every product references an issuer; orphan products are impossible
- [ ] Timestamps are stored with timezone
- [ ] Migration generated and applied cleanly against a fresh database
- [ ] Migration is reversible or the forward-only decision is documented
- [ ] Insert and select types exported for consumers
- [ ] A test proves a product round-trips through the real database

## Dependencies

TP-020, TP-022

## Component

db

## Risk

medium

## Estimate

3

clevervi:
The record a TrustPass ID resolves to. Closes #4 (TP-021).

## One acceptance criterion is superseded

> - [ ] TrustPass ID is the primary key

That predates [ADR 0005](docs/adr/0005-internal-keys-are-never-public.md), which
settled the opposite and answered the scalability question raised on #8. The key
is a `bigint identity` that never leaves the system; the identifier is a separate
uniquely indexed column.

The criterion existed to guarantee that one identifier resolves to one product.
That guarantee is intact — a unique index enforces it — without making a 128-bit
random value the clustering key. [Recorded on the issue](../../issues/4#issuecomment-5717857037)
rather than diverged from quietly.

**The issuer foreign key is RESTRICT, not CASCADE** — [product.ts:86](packages/db/src/schema/product.ts#L86)

Deleting an issuer that registered products would erase the provenance of
everything it signed, which is the one thing a passport exists to preserve.
Issuers get suspended, not deleted. Tested, and the code is `23001`
`restrict_violation` — not `23503`, which is what I asserted first. Postgres
raises RESTRICT immediately and NO ACTION at constraint-check time, so asserting
the wrong one would have passed while proving nothing.

**Status defaults to `draft`** — [product.ts:105](packages/db/src/schema/product.ts#L105)

A row existing is not a claim that the product is registered. Same reasoning that
makes `unverified` the issuer default.

**Category is an enum** — [product.ts:26](packages/db/src/schema/product.ts#L26)

Free text fragments one product across "GPU", "gpu" and "Graphics card", and that
fragmentation is permanent. Postgres adds an enum value cheaply and cannot remove
one, so the set is narrow and `other` exists because refusing to register a real
product is worse than an imprecise label.

**The identifier check is deliberately loose** — [product.ts:130](packages/db/src/schema/product.ts#L130)

`~ '^TP[0-9]'` and nothing more. ADR 0004 versions the format; a constraint
written today must not lock out a `TP2-` identifier issued later. A test asserts
that a future version is accepted.

## A defect this found

The `TrustPassId` brand was a module-local `unique symbol`. It rejected plain
strings correctly, but an unexported symbol cannot be named from outside the
module, so a consumer exporting a value containing one failed to compile:

```
error TS4023: Exported variable 'probe' has or is using name 'brand'
from external module .../trustpass-id but cannot be named.
```

Found by writing the code TP-023 will write — a value in `apps/api` holding a
product — rather than by reading the type. It would have blocked the registration
endpoint on its first export. Fixed with a named marker, and both halves verified
from a consumer package: exporting works, assigning a raw string still fails.

## Tests

70 pass. 21 new integration tests against real Postgres, covering the orphan
product, the RESTRICT delete, both enums, the shape and blank-serial checks, and
two issuers sharing a serial — which must stay legal, because stopping one issuer
registering the same serial twice is TP-024 and a different rule.

The SQLSTATE helper moved to `src/testing` so both suites share one definition
instead of two copies drifting.

## Verified from zero

Every earlier "applies cleanly to a fresh database" check in this project was
made with `DROP SCHEMA public CASCADE`, which is **not** a reset: Drizzle keeps
its applied-migration ledger in a separate `drizzle` schema that survives, so the
next migrate skips everything and leaves an empty database convinced it is
current. That is why the first attempt here failed.

`pnpm db:reset` removes the volume, and both migrations were run against a virgin
one:

```
Volume trustpass_postgres_data Created
Container trustpass-postgres Healthy
[✓] migrations applied successfully!
```

## Security impact

The RESTRICT foreign key is the meaningful one: provenance cannot be deleted out
from under a passport by removing its issuer.

Serials are stored whole and are **not** safe to display. Masking is TP-031, and
until it exists nothing should render this column publicly.

## Breaking changes

None. New table; the migration is additive.

---

- [ ] CI is green


packages/db/src/schema/product.ts
```diff
@@ -0,0 +1,147 @@
...
+     */
+    trustpassId: varchar("trustpass_id", { length: 64 }).$type<TrustPassId>().notNull(),
+
+    /**
+     * Who issued this product. Restrict rather than cascade: deleting an issuer
+     * with products would erase the provenance of everything it signed, which
+     * is the one thing a passport exists to preserve. Issuers are suspended,
+     * not deleted.
+     */
+    issuerId: bigint("issuer_id", { mode: "number" })
```
clevervi, line 88: **The decision I would most want a second opinion on.**

`RESTRICT` means an issuer with products can never be deleted — not by an admin, not by a GDPR-style erasure request, not by cleanup. That is deliberate: deleting the issuer erases the provenance of everything it signed, and provenance is the product.

But it is a commitment. If a legal request ever requires erasing an issuer, the answer will have to be anonymising the row, not deleting it. Worth knowing now rather than discovering it under a deadline.

packages/db/src/schema/product.ts
```diff
@@ -0,0 +1,147 @@
...
+ * shows and, later, what counts as an anomalous repair rate. Free text
+ * fragments the same product across "GPU", "gpu" and "Graphics card", and that
+ * fragmentation is permanent.
+ *
+ * Postgres can add an enum value cheaply and cannot remove one, so the initial
+ * set is deliberately narrow: the electronics this first market actually
+ * resells. `other` exists because refusing to register a real product is worse
+ * than an imprecise label.
+ */
+export const productCategory = pgEnum("product_category", [
```
clevervi, line 28: Postgres adds an enum value cheaply and **cannot remove one**. So this list is easy to grow and effectively permanent to shrink.

I kept it narrow — the electronics this first market actually resells — with `other` as the escape hatch, because refusing to register a real product is worse than an imprecise label.

The risk to watch: if `other` becomes the most common value, the enum is wrong and the fix is a migration plus a backfill.

packages/db/src/schema/product.ts
```diff
@@ -0,0 +1,147 @@
...
+     * Storage and display are separate concerns. A full serial shown publicly
+     * lets a third party impersonate the product elsewhere or claim against its
+     * warranty, so the passport masks it (TP-031). Masking at write time would
+     * destroy the value the system needs for verification.
+     */
+    serial: varchar("serial", { length: 120 }).notNull(),
+
+    category: productCategory("category").notNull(),
+
+    status: productStatus("status").notNull().default("draft"),
```
clevervi, line 108: `draft`, not `registered`. A row existing is not a claim that the product is registered — the same reasoning that makes `unverified` the issuer default.

Which transitions are legal is TP-025. Right now nothing stops a product going straight from `draft` to `retired`, or back from `suspended` to `active` without a trace. That gap is real until TP-025 closes it.

packages/db/src/schema/product.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * The manufacturer's serial, stored whole.
+     *
+     * Storage and display are separate concerns. A full serial shown publicly
+     * lets a third party impersonate the product elsewhere or claim against its
+     * warranty, so the passport masks it (TP-031). Masking at write time would
+     * destroy the value the system needs for verification.
+     */
+    serial: varchar("serial", { length: 120 }).notNull(),
```
clevervi, line 104: Stored whole, and **not safe to display**. A full serial shown publicly lets a third party impersonate the product elsewhere or claim against its warranty.

Masking is TP-031. Until it exists, nothing should render this column on a public surface — and there is currently no type or lint rule stopping that, only this comment.

packages/db/src/schema/product.ts
```diff
@@ -0,0 +1,147 @@
...
+    // how TP-024 will detect a duplicate. Lower-cased because serials get typed
+    // by hand off a sticker.
+    index("product_serial_idx").on(sql`lower(${table.serial})`),
+
+    // Shape only. The application parser in identity/trustpass-id.ts is
+    // authoritative, including the check symbol; this catches a raw UUID, an
+    // empty string or a truncated value reaching the column by another path.
+    // Deliberately loose about what follows "TP" so a future format version is
+    // not locked out by a constraint written today.
+    check("product_trustpass_id_shape", sql`${table.trustpassId} ~ '^TP[0-9]'`),
```
clevervi, line 137: Loose on purpose. ADR 0004 versions the identifier format, so a strict regex written today would reject a `TP2-` identifier issued later — and by then the constraint would be protecting nothing while blocking something real.

The application parser stays authoritative, including the check symbol. This catches a raw UUID, an empty string or a truncated value arriving by another path. A test asserts a future version is accepted.

clevervi (commented):
Went back over this before handing it to you. Nothing broken, but four decisions here are hard to reverse once real products exist, so they are flagged inline.

I also raised the risk label from medium to high. The issue was written before the design existed; this table is where product truth lives, and `CONTRIBUTING.md` puts that at high. So it waits for you rather than merging on green gates.

As always: I wrote this diff, so this is a second pass by the same judgement, not a second opinion.

PR #25, opened 2026-09-17
feat(identity): enforce the product status lifecycle

### LINKED ISSUE #5: [IDENTITY] TP-025 Model product status and its allowed transitions
## Objective

Model the product status lifecycle and the transitions allowed between states.

## Scope

**In scope:**

- Status enum: draft, registered, active, suspended, retired
- Allowed transitions, enforced rather than documented
- Tests over legal and illegal transitions

**Out of scope:**

- Lifecycle events, which are a separate append-only history (TP-050)
- Who is authorised to trigger a transition (TP-053)

## Context

Status is the product's current state. Lifecycle events are its history. Conflating them is a mistake that gets expensive later: status is mutable and singular, history is append-only and plural.

Transitions matter because some of them must be one-way. A suspended product that can quietly return to active erases the reason it was suspended. Suspension exists because something is wrong — a fraud flag, a theft report, a disputed claim — and clearing it should require a deliberate act that leaves a trace.

## Acceptance criteria

- [ ] Status stored as a database enum
- [ ] The allowed transition set is defined in one place, not scattered across call sites
- [ ] An illegal transition is rejected with an error naming both states
- [ ] Retired is terminal
- [ ] Suspended cannot return to active without an explicit reinstatement path
- [ ] Tests cover every legal transition and a representative set of illegal ones
- [ ] The transition rules are documented where a reader will find them

## Dependencies

TP-021

## Component

db

## Risk

high — status gates what the public passport is allowed to claim.

## Estimate

3

clevervi:
Closes #5 (TP-025).

The issue asked for transitions **enforced rather than documented**, so they are
enforced in the database. Application code is bypassed by the first write path
that does not go through it, and status gates what the public passport is allowed
to claim.

[`docs/product-lifecycle.md`](docs/product-lifecycle.md) has the diagram and the
full table.

## Two rules that are deliberately inconvenient

**`suspended` cannot return to `active` in one step.** Suspension exists because
something is wrong — a fraud flag, a theft report, a disputed claim. A product
that quietly becomes active again erases the reason it was ever suspended, and
the next buyer sees no trace of the episode. Clearing a suspension returns it to
`registered`; activating it again is a separate act.

**`retired` is terminal.** A product that can come back from end of life makes
the word meaningless on a passport.

## Beyond the acceptance criteria

Creation is restricted to `draft` and `registered`. The criteria do not mention
inserts, but a transition guard with an open insert path enforces nothing — the
whole table is bypassed by inserting the end state, and a product born `active`
was never registered by anyone.

## The part worth reviewing

The rules exist **twice**: in TypeScript for typed errors the API can report, and
in a trigger no write can sidestep. Two copies drift. There is no way to merge
them into one artefact, so instead:

`product-status.integration.test.ts` walks **all 25** source-to-target pairs and
asserts the database agrees with the TypeScript table. A divergence is a red test
rather than a silent hole in the guarantee.

That assertion was confirmed to fail. Adding `suspended → active` to the
TypeScript table alone:

```
× suspended to active
Tests  3 failed | 124 passed (127)
```

Restored, 127 pass. A drift detector that has never detected drift has not been
shown to detect anything.

## Implementation notes

`TP001` is a project-defined SQLSTATE, so an illegal move is distinguishable from
an ordinary check violation — "that move is not allowed" is a different answer
from "the database said no". Verified against Postgres before the trigger was
written.

The trigger is `BEFORE UPDATE OF status`, so ordinary writes that leave status
out of the `SET` list never fire it, and setting a status to the value it already
holds returns early. Otherwise every brand correction would be treated as a
transition — and the test suite covers both.

## Tests

139 pass. 71 new: 46 unit over the transition table, 25 integration pairs against
the real trigger, plus insert restrictions, ordinary-write pass-through, and the
error message naming both states.

One correction along the way: the trigger's message lives on the driver error,
not on Drizzle's wrapper, which stringifies as `Failed query: ...` and drops it.
Asserting against the surfaced error tested the wrapper. There is now a
`sqlMessageOf` helper that walks the cause chain, mirroring `sqlStateOf`.

Biome also caught a genuine footgun: the test's route table used a `then`
property, which makes an object thenable and would misbehave under `await`.
Renamed.

Verified from a virgin volume with `pnpm db:reset`, so all three migrations apply
in order to an empty database.

## Security impact

This is the enforcement that makes `status` trustworthy, and the passport reads
`status` to decide what it may claim. Before this, any code path could move a
product from `retired` to `active`, or clear a suspension silently.

Still open, and named in the doc: nothing records *why* a status changed. That is
TP-050. And *who* may change it is TP-053 — the trigger enforces which moves are
legal, not who may make them.

## Breaking changes

None. New trigger on an empty table.

---

- [ ] CI is green


clevervi (commented):
Second pass before handing it over. The design has one deliberate compromise and two gaps that are real until later tickets close them; all three are inline.

Also worth saying: I started this work on `develop` instead of a branch. Caught it before committing, so nothing was lost, and branch protection would have rejected the push anyway. The rule earned its keep.

packages/db/src/domain/product-status.ts
```diff
@@ -0,0 +1,89 @@
...
+ *   one a buyer sees in a resale. It can be flagged or retired.
+ * - `suspended` exists because something is wrong: a fraud flag, a theft report,
+ *   a disputed claim. It deliberately cannot return to `active` in one step.
+ *   Clearing a suspension returns the product to `registered`, and activating it
+ *   again is a separate, separately recorded act. A product that quietly becomes
+ *   active again erases the reason it was ever suspended.
+ * - `retired` is terminal. A product that can come back from end-of-life makes
+ *   "retired" mean nothing on a passport.
+ */
+export const PRODUCT_STATUS_TRANSITIONS: Readonly<Record<ProductStatus, readonly ProductStatus[]>> =
```
clevervi, line 32: **The compromise I would most want challenged.**

These rules exist twice — here, and in the trigger in `0002_product_status_transition_guard.sql`. Duplication is normally a defect, and I chose it knowingly: application code gives typed errors the API can report, and a trigger cannot be sidestepped by a write path that forgets to call this.

There is no artefact that can be both. So the mitigation is a test that walks all 25 pairs and fails if the two disagree, and I confirmed it fails when they do.

If you would rather have one copy, the honest alternative is dropping the TypeScript table and letting every caller handle `TP001` — which costs the typed error and makes the API's job harder. I think this trade is right, but it is a trade.

packages/db/src/domain/product-status.ts
```diff
@@ -0,0 +1,89 @@
...
+  {
+    draft: ["registered", "retired"],
+    registered: ["active", "suspended", "retired"],
+    active: ["suspended", "retired"],
+    suspended: ["registered", "retired"],
```
clevervi, line 37: `suspended` goes to `registered`, deliberately not to `active`.

The issue asked that reinstatement be explicit. This is the simplest reading that is enforceable identically in TypeScript and SQL: clearing a suspension and activating again are two acts, so a product cannot silently return to circulation.

The cost is real — reinstating a wrongly-suspended product takes two operations, and until TP-050 exists neither leaves a record of why. If that turns out to be too awkward in practice, the alternative is allowing `suspended → active` only through a function that demands a reason, which the trigger cannot mirror.

packages/db/drizzle/0002_product_status_transition_guard.sql
```diff
@@ -0,0 +1,68 @@
...
+--
+-- SQLSTATE TP001 is a project-defined code. It distinguishes an illegal
+-- transition from an ordinary check violation, so a caller can report "that
+-- move is not allowed" rather than "the database said no".
+
+CREATE OR REPLACE FUNCTION product_status_guard() RETURNS trigger
+LANGUAGE plpgsql
+AS $$
+BEGIN
+  IF TG_OP = 'INSERT' THEN
```
clevervi, line 21: Not in the acceptance criteria, added anyway.

Without it the transition table is bypassed completely: `INSERT ... status = 'active'` produces a product that was never registered by anyone, and the guard below never runs. A transition guard with an open insert path enforces nothing.

packages/db/drizzle/0002_product_status_transition_guard.sql
```diff
@@ -0,0 +1,68 @@
...
+      RAISE EXCEPTION 'A product cannot be created with status "%".', NEW.status
+        USING ERRCODE = 'TP001';
+    END IF;
+
+    RETURN NEW;
+  END IF;
+
+  -- Updating a brand or a serial leaves the status alone. Treating that as a
+  -- transition would block every ordinary write.
+  IF NEW.status = OLD.status THEN
```
clevervi, line 35: Easy to get wrong, and expensive if you do.

Without this early return, `UPDATE product SET status = status, brand = 'Gigabyte'` would be treated as a transition. Combined with `BEFORE UPDATE OF status` on the trigger, ordinary writes that never mention status do not fire it at all. Both paths are covered by tests.

packages/db/src/domain/product-status.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * True when the move is permitted.
+ *
+ * Staying in the same status is always permitted: an update that changes a
+ * brand or a serial leaves the status alone, and treating that as a transition
+ * would block every ordinary write.
+ */
+export function canTransition(from: ProductStatus, to: ProductStatus): boolean {
+  if (from === to) {
```
clevervi, line 67: **Two gaps that are real until later tickets close them**, both named in `docs/product-lifecycle.md`:

Nothing records *why* a status changed. A product can go `active → suspended` and the row shows only the new value — no reason, no timestamp beyond `updated_at`, no actor. That is TP-050.

Nothing restricts *who* may change it. The trigger enforces which moves are legal, not who may make them. That is TP-053.

So right now a suspension is enforceable but unexplainable, which is fine for a schema and would not be fine in production.

PR #26, opened 2026-09-17
feat(api): register a product via POST /products

### LINKED ISSUE #6: [IDENTITY] TP-023 Register a product via POST /products
## Objective

Expose `POST /products` so an issuer can register a product and receive its TrustPass ID.

## Scope

**In scope:**

- `POST /products` with Zod request and response schemas
- TrustPass ID generation on registration
- Persistence
- OpenAPI documentation generated from the same schemas
- Structured error responses

**Out of scope:**

- Authentication (the issuer is supplied explicitly until TP-141)
- Bulk CSV import (TP-153)
- Lifecycle event emission (TP-051)

## Context

This is the first write endpoint and it sets the pattern every later one copies: validation at the boundary, generated ID, persisted record, documented contract, structured errors.

Registration is not idempotent by nature but it will be retried by clients on timeout. A retry that silently creates a second identity for the same physical product is a correctness bug that surfaces months later, on the resale market, as two passports for one GPU.

## Acceptance criteria

- [ ] `POST /products` accepts brand, model, serial, category and issuer
- [ ] Request validated at the boundary; malformed input returns 422 with the offending fields
- [ ] A TrustPass ID is generated server side and never accepted from the client
- [ ] Response returns the created product including its TrustPass ID
- [ ] Endpoint appears in `/openapi.json` with request and response schemas
- [ ] Unknown issuer returns a structured error, not a database constraint message
- [ ] Errors follow one documented shape across the endpoint
- [ ] Tests cover the happy path, validation failure, and unknown issuer
- [ ] At least one test runs against the real database

## Dependencies

TP-020, TP-021, TP-022

## Component

api

## Risk

medium

## Estimate

5

clevervi:
Closes #6 (TP-023). The first write endpoint, and the pattern every later one
copies.

## A derivation the issue could not have made

The issue says "the issuer is supplied explicitly until TP-141". It cannot be
supplied as an **id**: [ADR 0005](docs/adr/0005-internal-keys-are-never-public.md)
keeps the internal key inside the system, and
[ADR 0006](docs/adr/0006-issuers-are-identified-by-registration-number.md) makes
an issuer's public identity its registration number.

So the body carries country and registration number, and the service resolves
them:

```json
{
  "issuer": { "country": "CO", "registrationNumber": "900123456-7" },
  "brand": "ASUS",
  "model": "ROG Strix RTX 5070 Ti",
  "serial": "M1LMCS004896",
  "category": "gpu"
}
```

**422 for validation, not Hono's default 400** — [app.ts:15](apps/api/src/app.ts#L15)

Through a `defaultHook`. Without it, the documented error envelope would apply to
every response except the one a caller is most likely to hit.

**422 for an unknown issuer, not 404** — [products.ts:108](apps/api/src/routes/products.ts#L108)

The endpoint exists; the instructions in the body cannot be carried out. A 404
on `POST /products` reads as "that endpoint does not exist".

**The API cannot build a query** — [issuer-repository.ts](packages/db/src/repositories/issuer-repository.ts)

Importing `drizzle-orm` into `apps/api` would let any route write arbitrary SQL.
ADR 0005 says the internal key must never cross the boundary and notes the type
system does not enforce it. Two thin repositories keep query construction inside
`packages/db`, which makes the rule structural: the API cannot select a column it
should not return if it cannot write the select. `drizzle-orm` is a
**devDependency** of `apps/api`, used only by an integration test that verifies
storage.

**The TrustPass ID is never read from the request** — [register-product.ts:74](apps/api/src/products/register-product.ts#L74)

A caller choosing its own identifier could collide with one already printed on a
label. A test asserts a client-supplied `trustpassId` is ignored, not honoured.

## Known gap, and it has a test

**Nothing stops one issuer registering the same serial twice.** A client retrying
after a timeout produces two passports for one physical GPU — the exact failure
the system exists to prevent.

That is TP-024, and it belongs in the database, because two concurrent retries
both pass an application check before either writes.

An integration test asserts the behaviour **as it is**, with a comment saying so.
TP-024 has to break that test to land, which is the point of writing it.

## Tests

158 pass. 27 new: HTTP contract with the service stubbed, and storage verified
against real Postgres.

Worth noting three:

- The response is asserted to contain no `id` and no `issuerId`. ADR 0005 is a
  review concern with nothing enforcing it; this enforces it.
- The unknown-issuer message is asserted **not** to match
  `/constraint|violates|relation|SQLSTATE/i`.
- `registerProduct` is asserted not to be called at all when the body is invalid.

## Verified over real HTTP

Not only through `app.request`. Against the built artefact:

```
--- happy path: HTTP 201
{ "trustpassId": "TP1-7V5YJY4GCNZ88QV504ZNGG33DMU", "status": "registered",
  "issuer": { "registrationNumber": "900123456-7", "verificationStatus": "unverified" } }

--- malformed: HTTP 422   (five offending fields named)
--- unknown issuer: HTTP 422  { "error": "issuer_not_found" }
```

## Security impact

- The internal key does not appear in any response, asserted rather than assumed.
- Database errors never reach the caller; an unknown issuer is a domain outcome,
  not a caught constraint violation.
- The full serial **is** returned — to the issuer that submitted it, which is
  its own data. The public passport masks it, and that is TP-031.
- No authentication yet. Anyone who can reach the endpoint can register against
  any issuer. That is TP-141, and until it lands this must not be exposed
  publicly.

## Breaking changes

None. New endpoint.

---

- [ ] CI is green
- [x] `pnpm-lock.yaml` is committed — `drizzle-orm` added as a devDependency


apps/api/src/products/register-product.integration.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    // reports who said it, not whether it is true. The response has to make
+    // that visible rather than implying verification happened.
+    const created = (await (await post(body({ serial: `${run}-UNVERIFIED` }))).json()) as {
+      issuer: { verificationStatus: string };
+    };
+
+    expect(created.issuer.verificationStatus).toBe("unverified");
+  });
+
+  it("currently allows one issuer to register the same serial twice", async () => {
```
clevervi, line 127: **A test that asserts a bug.**

Registering the same serial twice currently succeeds. A client retrying after a timeout gets two passports for one physical GPU — the exact failure this system exists to prevent.

I could have left it undocumented and let TP-024 discover it. Instead this asserts the behaviour as it is, so TP-024 **has to break this test to land**. A known gap with no test is a gap that gets forgotten.

If you would rather not have a green test describing wrong behaviour, the alternative is `it.fails` or a skipped test with a TODO — both weaker, because neither fails when the bug is fixed.

apps/api/src/app.ts
```diff
@@ -3,30 +3,44 @@ import { cors } from "hono/cors";
...
+import { registerProductRoutes } from "./routes/products.js";
 
 export function createApp(deps: AppDependencies): OpenAPIHono {
-  const app = new OpenAPIHono();
+  const app = new OpenAPIHono({
+    // Without this, a schema failure returns Hono's own 400 body and the
+    // documented error envelope applies to every response except the one a
+    // caller is most likely to hit. 422 rather than 400: the JSON parsed fine,
+    // the instructions in it could not be followed.
+    defaultHook: (result, c) => {
```
clevervi, line 16: This one hook decides the error shape for the whole API.

Without it, Hono returns its own 400 body for schema failures, and the documented envelope would cover every response except the one callers hit most. 422 rather than 400 because the JSON parsed fine — the instructions in it could not be followed.

Worth agreeing on now: every later endpoint inherits this, and changing it after clients exist is a breaking change.

apps/api/src/routes/products.ts
```diff
@@ -0,0 +1,120 @@
...
+    const body = c.req.valid("json");
+    const result = await deps.registerProduct(body);
+
+    if (!result.ok) {
+      // A well-formed request naming something that does not exist is 422, not
+      // 404: the endpoint is there, the instructions in the body cannot be
+      // carried out. A 404 here would read as "POST /products does not exist".
+      return c.json(
+        {
+          error: ApiErrorCode.ISSUER_NOT_FOUND,
```
clevervi, line 111: 422 rather than 404 for an unknown issuer.

The endpoint exists; the body names something that does not. A 404 on `POST /products` reads as "that endpoint is not there", which sends a caller debugging the wrong thing.

Defensible either way — plenty of APIs use 404 for a missing referenced resource. I would rather be wrong consistently than inconsistently, so if you prefer 404, now is the time.

apps/api/src/products/register-product.ts
```diff
@@ -0,0 +1,114 @@
...
+    input.issuer.country,
+    input.issuer.registrationNumber,
+  );
+
+  if (!issuer) {
+    return { ok: false, reason: "issuer_not_found" };
+  }
+
+  const created = await insertProduct(db, {
+    trustpassId: generateTrustPassId(),
```
clevervi, line 87: Generated here, never read from the request.

A caller that could choose its own identifier could collide with one already printed on a label, or reuse a value seen on someone else's product. A test asserts a client-supplied `trustpassId` is dropped rather than honoured — the body schema simply has no such field, so Zod strips it.

Also note the status: `registered`, not `draft`. Calling this endpoint is the issuer committing to the record.

packages/db/src/repositories/issuer-repository.ts
```diff
@@ -0,0 +1,24 @@
+import { and, eq } from "drizzle-orm";
+import type { Database } from "../client.js";
+import { type Issuer, issuer } from "../schema/issuer.js";
+
+/**
+ * Finds an issuer by the identity the outside world can name it with.
+ *
+ * Per ADR 0006 that is the registration number its national authority
```
clevervi, line 8: The reason these repositories exist at all.

The alternative was importing `drizzle-orm` into `apps/api`, which is one line and lets any route write arbitrary SQL. ADR 0005 says the internal key must never cross the API boundary and admits nothing enforces that. Keeping query construction in this package makes it structural: the API cannot select a column it should not return if it cannot write the select.

The cost is indirection for two functions. I think it is worth it here because the rule it protects is a trust boundary, not a style preference.

Note there is deliberately no `findById` — an id is not a valid lookup key from outside.

clevervi (commented):
Second pass before handing over. This is the first write endpoint, so its shape becomes the default for every later one — the things worth arguing about are inline.

One gap ships knowingly and has a test asserting it. One risk is not a gap but a deployment constraint, and it is the reason this is labelled high.

PR #27, opened 2026-09-17
feat: one live identity per serial, per issuer

### LINKED ISSUE #7: [IDENTITY] TP-024 Prevent duplicate active identities for one serial
## Objective

Guarantee that the same issuer cannot create two active identities for the same serial.

## Scope

**In scope:**

- Database constraint enforcing the rule
- API-level handling that turns the violation into a clear response
- Tests including concurrent registration

**Out of scope:**

- Detecting the same serial registered by *different* issuers — that is a fraud signal, not a constraint (TP-111)

## Context

This is the rule that makes the whole system mean anything. If one physical product can hold two active passports, then a passport proves nothing: a seller picks whichever history looks better and shows that one.

The constraint has to live in the database. An application-level check reads, decides, and then writes, and two concurrent requests can both pass the read before either writes. Under retry storms that is not a rare race — it is the expected behaviour.

"Active" is the operative word. A retired product's serial should not permanently poison the serial space; a replacement unit under warranty legitimately reuses it.

## Acceptance criteria

- [ ] A partial unique index enforces uniqueness over (issuer, serial) for active identities only
- [ ] A duplicate registration returns a clear, documented error, never a raw database message
- [ ] The error names the conflict and references the existing TrustPass ID
- [ ] A test fires concurrent registrations of the same serial and asserts exactly one succeeds
- [ ] A test proves a serial can be reused once the prior identity is retired
- [ ] The definition of "active" is documented in the schema

## Dependencies

TP-021, TP-023, TP-025

## Component

db

## Risk

critical — without this, a passport is not evidence of anything.

## Estimate

3

clevervi:
Closes #7 (TP-024). The last item in v0.2.0, and the rule that makes a passport
mean anything: if one physical product can hold two live identities, a seller
shows whichever history looks better and the passport proves nothing.

## Why it is in the database

An application check reads, decides, then writes. Two concurrent retries both
pass the read before either writes — under a retry storm that is not a rare race,
it is the expected behaviour.

A partial unique index on `(issuer_id, lower(serial)) WHERE status <> 'retired'`.

**Proven, not assumed.** Five concurrent registrations of one serial, over real
HTTP:

```
five concurrent: {"201": 1, "409": 4}
```

And the tests genuinely depend on it. Dropping the index turns 4 of the 5
duplicate tests red; restoring it returns all 162 to green.

## Three decisions

**Lower-cased** — [product.ts:139](packages/db/src/schema/product.ts#L139)

The lesson the issuer table taught. A plain unique index is case sensitive, so
`M1LMCS004896` and `m1lmcs004896` would both be accepted and the constraint would
stop nothing anyone actually types. That exact defect shipped in the issuer PR
before it was caught.

**"Live" means every status except `retired`**

A draft, registered, active or suspended product still holds an identity. But a
retired product must not poison its serial forever — a warranty replacement unit
legitimately carries the serial of the unit it replaced.

So **retiring a product is the act that frees its serial**, which is a second
consequence of retirement worth knowing. Now recorded in
[`docs/product-lifecycle.md`](docs/product-lifecycle.md).

**Scoped to one issuer**

Serials are unique only within a manufacturer's own numbering, so two importers
holding colliding serials is legitimate. The same serial across *different*
issuers is a fraud signal to weigh later (TP-111), not a constraint to enforce
now. A test asserts it stays allowed.

## 409, and what it carries

```json
{
  "error": "duplicate_serial",
  "message": "Serial OLK8E-DUP already belongs to TP1-3JY3SD5DXWD6HDB6EMNWRXSA1M~ under this issuer. Retire that identity before registering the serial again."
}
```

409 rather than 422: the request is valid and would have been accepted a moment
earlier. It conflicts with state that exists.

The existing TrustPass ID is returned deliberately. A client that timed out
mid-registration and retried learns what it already created rather than only
being told no — otherwise a retry has no recovery except guessing. The message is
asserted **not** to match `/constraint|violates|relation|SQLSTATE|index/i`.

`insertProduct` translates only this one constraint. Any other database failure
rethrows, because swallowing an unrecognised error would turn a real fault into a
plausible-looking rejection a user would be told was their own mistake.

## A test that had to break

TP-023 shipped a test asserting that a duplicate **succeeds**, with a comment
saying TP-024 would have to break it to land. It did, and it is replaced.

## A defect the index exposed

Every product built by the product table suite shared one serial — the counter
was never incremented and nothing complained. Those fixtures had been creating
duplicates of each other, unnoticed, since TP-021.

## Tests

162 pass. 6 new covering the duplicate, the case variant, reuse after
retirement, the concurrent race, and the cross-issuer case that must stay legal.

## A correction

I first reported this as broken over HTTP — a 500 on the duplicate. It was my
harness: `TaskStop` killed the pnpm wrapper but not its node child, so requests
hit the **previous** build from TP-023, which had no duplicate handling. The log
said `EADDRINUSE` and the new server never started. Killed the stale process,
re-ran, and the results above are from the current artefact.

## Security impact

This is the integrity constraint the whole trust model rests on. Without it a
passport is not evidence of anything.

Still absent: no authentication, so anyone reaching the endpoint can register
against any issuer (TP-141). Until that lands the API must not be exposed
publicly.

## Breaking changes

None functionally — but a duplicate registration that previously returned 201 now
returns 409. No client exists yet.

---

- [ ] CI is green


packages/db/src/schema/product.ts
```diff
@@ -124,11 +124,38 @@ export const product = pgTable(
...
+    // or suspended product still holds a live identity and must block a second
+    // one. Retirement releases the serial, because a warranty replacement unit
+    // legitimately carries the serial of the unit it replaced, and a retired
+    // product poisoning its serial forever would make that impossible.
+    //
+    // Scoped to one issuer on purpose. Two issuers holding units with colliding
+    // serials is legitimate -- serials are only unique within a manufacturer's
+    // own numbering. The same serial across different issuers is a fraud signal
+    // to weigh, not a constraint to enforce (TP-111).
+    uniqueIndex("product_live_issuer_serial_idx")
```
clevervi, line 155: **`lower(serial)` is the carried lesson, not a flourish.**

The issuer table shipped this exact constraint case-sensitive, and I only caught it by querying the database instead of reasoning about it. Two rows went in where one should have. Serials are typed by hand off a sticker, so the same defect here would be worse.

**"Live" means every status except `retired`**, and that has a consequence worth agreeing on: retiring a product is now the act that *frees its serial*. A warranty replacement unit legitimately carries the serial of the unit it replaced, so a retired product must not poison the serial forever. Recorded in `docs/product-lifecycle.md` rather than left implicit in a WHERE clause.

packages/db/src/repositories/product-repository.ts
```diff
@@ -1,21 +1,103 @@
...
 
-  if (!created) {
-    // A successful insert always returns its row. Reaching here means something
-    // changed underneath us, not a case a caller can handle.
-    throw new Error("Product insert returned no row.");
+    if (violation.code === UNIQUE_VIOLATION && violation.constraint === LIVE_SERIAL_INDEX) {
+      return { ok: false, reason: "duplicate_live_serial" };
+    }
+
+    throw error;
```
clevervi, line 71: **This rethrow is the important line in the file.**

Only one constraint is translated into an outcome. Everything else propagates.

The tempting version catches any `23505` — or worse, any error — and reports "duplicate serial". That would turn a genuine fault into a plausible-looking rejection, and the caller would tell a user they made a mistake they did not make. A wrong error message that looks right is worse than a stack trace.

apps/api/src/products/register-product.ts
```diff
@@ -93,6 +104,21 @@ export async function registerProduct(
...
+  if (!inserted.ok) {
+    const existing = await findLiveProductBySerial(db, issuer.id, input.serial);
+
+    return {
+      ok: false,
+      reason: "duplicate_serial",
+      // The lookup can only miss if the winning row was retired between the
+      // rejection and this read. Reporting the conflict without an identifier
+      // is still the honest answer; inventing one would not be.
+      existingTrustpassId: existing?.trustpassId ?? "unknown",
```
clevervi, line 116: **The edge I am least happy with.**

The lookup can only miss if the winning row is retired between the index rejecting our insert and this read — a very small window, but not zero.

I chose to report the conflict with `"unknown"` rather than invent an identifier or swallow the error. Reporting a conflict without an identifier is still true; anything else would not be.

The alternative is doing both in one transaction so the read cannot race. That is more correct and more machinery, and I did not think a window this narrow justified it for a first version. Worth disagreeing with.

apps/api/src/products/register-product.integration.test.ts
```diff
@@ -124,26 +124,96 @@ describe.skipIf(!databaseUrl)("POST /products against a real database", () => {
...
+        .where(eq(schema.product.trustpassId, original.trustpassId as never));
+
+      const replacement = await post(body({ serial }));
+      expect(replacement.status).toBe(201);
+
+      const rows = await db.select().from(schema.product).where(eq(schema.product.serial, serial));
+      expect(rows).toHaveLength(2);
+    });
+
+    it("lets exactly one of several concurrent registrations win", async () => {
```
clevervi, line 179: This is the test that justifies putting the rule in the database rather than in a service.

Five simultaneous registrations of one serial. Every one of them would pass an application-level "does this exist?" check before any of them wrote. Exactly one row survives.

It is also load-bearing: dropping the index turns this and three sibling tests red, which I checked rather than assumed.

apps/api/src/routes/products.ts
```diff
@@ -102,6 +107,22 @@ export function registerProductRoutes(app: OpenAPIHono, deps: AppDependencies):
     const body = c.req.valid("json");
     const result = await deps.registerProduct(body);
 
+    if (!result.ok && result.reason === "duplicate_serial") {
+      // 409, not 422: the request is entirely valid and would have been
```
clevervi, line 111: **What is still missing, and it gates deployment.**

There is no authentication. Anyone who can reach this endpoint can register products against *any* issuer, including one they do not own, and now also learn — via the 409 message — the TrustPass ID of a product they did not register, by guessing a serial.

That last part is new with this change and worth stating plainly: the recovery path this adds for honest clients is also an oracle for a dishonest one. It is the right trade while the API is unauthenticated and unexposed, and it stops being the right trade the moment TP-141 is not done first.

**This must not be exposed publicly before TP-141.**

clevervi (commented):
Last item in v0.2.0, and labelled critical for a reason: without it a passport is not evidence of anything. Five things inline — one carried lesson, one edge case I am not fully happy with, and the reason nothing here should be exposed publicly yet.

Issue #33, opened 2026-09-17
[PASSPORT] TP-033 State each verification claim separately


clevervi:
## Objective

State what TrustPass has verified and what it has not, claim by claim.

## Scope

**In scope:**

- A pure function computing the verification claims from product and issuer state
- The verification panel on the passport page
- Prominence for the two states that invert the page's meaning

**Out of scope:**

- Secure NFC tag verification (TP-130+)
- Warranty verification (TP-040+)
- The verification service and its endpoint (TP-070, TP-072)

## Context

**This is the milestone, and the project's entire argument.**

Per ADR 0003, TrustPass proves **identity**, not physical **authenticity**. If a fraudulent issuer registers a counterfeit as genuine, the system faithfully records that the issuer said so. It has inspected nothing.

The interface must never render an unqualified "authentic" badge. What can honestly be asserted today:

| Claim | State | Backed by |
| --- | --- | --- |
| Issuer | its real verification status | `issuer.verification_status` |
| Serial | `recorded` — **not** "verified" | the issuer supplied it; nothing checked it against the object |
| Secure tag | `not_present` | no NFC exists yet |
| Warranty | `not_recorded` | no warranty table exists yet |
| Physical product | `not_verifiable` | permanent — a ledger cannot inspect an object |

The second row is where this either tells the truth or becomes a green checkmark. "Serial: VERIFIED" would be a lie: the serial was *recorded*, not checked.

Two states must be prominent rather than buried, because they invert what the page means:

- An **unverified issuer** — the product carries exactly the weight of an unchecked company's word
- A **suspended product** — something is wrong with this specific unit

**The API returns `{ claim, state }` codes; the page owns the wording.** Machine consumers need stable codes, not prose, and TP-072 will reuse this function rather than reimplement it.

## Acceptance criteria

- [ ] Claims are computed by a pure function, unit-tested
- [ ] Each claim is displayed separately and never collapsed into a single verdict
- [ ] No state, in any combination, renders as an unqualified "authentic"
- [ ] The serial claim reads as recorded, never as verified
- [ ] Physical authenticity is always shown as not independently verified
- [ ] An unverified issuer is prominent, not buried in a list
- [ ] A suspended product is prominent
- [ ] The API returns stable codes; human wording lives in the page
- [ ] A test asserts that no combination of inputs produces a bare "verified" verdict for the product as a whole

## Dependencies

TP-030, TP-022

## Component

api

## Risk

[...]

Issue #32, opened 2026-09-17
[PASSPORT] TP-031 Show the product identity, with the serial masked


clevervi:
## Objective

Show what the product is: brand, model, serial, category, status, and who issued it.

## Scope

**In scope:**

- The identity panel on the passport page
- Serial masking
- Issuer identity, including the registration number

**Out of scope:**

- The verification claims (TP-033)
- Lifecycle history (TP-034, deferred to v0.5.0)

## Context

The serial is the manufacturer's identifier and it is sensitive. A full serial published on a page anyone can scan lets a third party impersonate the product elsewhere or claim against its warranty. It must be masked.

**Masking has an edge case that would leak everything.** A serial may legally be as short as two characters — `product_serial_not_blank` requires `length(trim(serial)) >= 2` and the request schema allows `min(2)`. A naive "show the last four" discloses the entire serial for anything four characters or shorter.

The issuer's registration number **is** shown. It is public data, national registries publish it, and it lets a buyer verify the issuer independently rather than taking TrustPass's word for it. That is the point of ADR 0003.

## Acceptance criteria

- [ ] Brand, model, category, status and issuer are displayed
- [ ] The serial is masked to its last four characters
- [ ] The last four are revealed **only** when the serial is at least eight characters; below that nothing is revealed
- [ ] The masking function is pure and unit-tested, including the short-serial case
- [ ] The issuer's company name, country and registration number are shown
- [ ] A `suspended` product is visibly marked, not shown as an ordinary passport
- [ ] The full serial never appears in the page source or the API response

## Dependencies

TP-030

## Component

web

## Risk

high — the masking is a security control, and its edge case discloses the value it exists to protect.

## Estimate

3

clevervi:
## One criterion corrected before implementation

> The last four are revealed **only** when the serial is at least eight characters

At exactly eight characters, "the last four" is **half** the serial. The same plan that set this threshold also said "masking that discloses half the value is not masking" — the two contradicted each other.

The implementation states the rule as the principle instead of a number: **the last four are revealed only when more characters stay hidden than are shown.** The first length that satisfies it is nine.

| Length | Revealed |
| --- | --- |
| 2–8 | nothing |
| 9 | last 4, 5 hidden |
| 12 | last 4, 8 hidden |

Real serials are unaffected — GPU, laptop and phone serials run 10 characters and up. A mutation check confirms the guard: removing it turns six tests red, including "never returns the whole serial" for a two-character serial.

Issue #31, opened 2026-09-17
[PASSPORT] TP-030 Resolve a TrustPass ID to a public passport


clevervi:
## Objective

Make a registered product readable from outside. A TrustPass ID entered or scanned resolves to a page.

## Scope

**In scope:**

- Repository lookup of a product by its TrustPass ID, joined with its issuer
- `GET /products/{trustPassId}` returning a public projection
- The page at `/trustpass/[trustPassId]` resolving an identifier
- Distinguishing a mistyped identifier from an unknown one

**Out of scope:**

- What the page displays in detail (TP-031, TP-033)
- The QR code (TP-032)
- Authentication (TP-141)

## Context

This is the first public read surface in the system. Everything registered in v0.2.0 is currently invisible from outside.

Two constraints carry over and are not negotiable here.

Per ADR 0005 the internal key never crosses the API boundary. The projection must exclude `product.id`, `product.issuerId` and `issuer.id`. Nothing in the type system enforces this, so it is asserted in tests.

Per ADR 0004 the TrustPass ID carries a check symbol, and this is where that machinery first earns its keep. A mistyped identifier and an unregistered product are **different answers**. Telling someone their product is unregistered when they fat-fingered one character is a false fraud accusation made to someone deciding whether to complete a purchase.

The web app must not query Postgres directly. The boundary established in v0.2.0 — `apps/api` cannot build a query, so the internal key cannot leak — dissolves the moment the page opens its own connection.

## Acceptance criteria

- [ ] Repository function looks a product up by TrustPass ID and returns its issuer in the same query
- [ ] `GET /products/{trustPassId}` returns 200 with a projection containing no internal keys
- [ ] A well-formed identifier with no matching product returns 404
- [ ] An identifier failing the check symbol returns 422, not 404
- [ ] The endpoint appears in `/openapi.json` with its response schema
- [ ] A database error never reaches the caller as a raw message
- [ ] The page renders when the identifier resolves, and states the difference between unknown and mistyped
- [ ] The page renders an honest error when the API is unreachable, not a blank passport
- [ ] A test asserts the response contains no `id` or `issuerId`
- [ ] At least one test runs against the real database

## Dependencies

TP-020, TP-021, TP-022

## Component

api

## Risk

high — the first public read surface, and where an internal key would leak.

## Estimate

8

clevervi:
## Two criteria refined by the design review, before any code

[...]

PR #36, opened 2026-09-17
feat: read a public passport by TrustPass ID

clevervi:
PR 1 of 3 for v0.3.0. The data and API half of the passport: everything the page in PR 2 will render. No UI here.

Refs #31 (TP-030), #32 (TP-031), #33 (TP-033). They close in PR 2, when the page makes their criteria true — each has criteria spanning API and page.

## What it adds

`GET /passports/{trustpassId}` — the first public read surface. Everything registered in v0.2.0 was invisible from outside.

```
GET /passports/TP1-4C6V1TSVSEH70HNMC2N0WMSD3H8   → 200, no-store
{
  "serial": { "suffix": "4896", "hiddenCharacters": 27 },
  "registeredOn": "2026-09-17",
  "issuer": { "registrationNumber": "900123456-7", "verificationStatus": "verified" },
  "claims": [
    { "claim": "issuer", "state": "verified" },
    { "claim": "serial", "state": "recorded" },
    { "claim": "secure_tag", "state": "not_present" },
    { "claim": "warranty", "state": "not_recorded" },
    { "claim": "physical_authenticity", "state": "not_verifiable" }
  ]
}
```

Verified over HTTP against the built artefact — whole serial absent, `id` absent, `issuerId` absent.

**`/passports`, not `/products`** — `POST /products` returns the full serial; the same prefix returning a masked one is an asymmetry that invites a mistake. And when TP-141 adds auth, `/products` locks and `/passports` stays public: a prefix rule, not a per-method exception. Recorded on #31.

**A typo is not an unknown product.**

| Input | Status | Code |
| --- | --- | --- |
| check symbol fails | 422 | `mistyped_trustpass_id` |
| otherwise unparseable | 422 | `malformed_trustpass_id` |
| no row, or a draft | 404 | `passport_not_found` |

**The path-param schema is deliberately permissive** — [passports.ts:14](apps/api/src/routes/passports.ts#L14). This is the easiest thing in the milestone to break by "improving" it. A format regex routes every bad ID through `defaultHook` into `validation_failed`, erasing the distinction above — and also rejects lower case, breaking ADR 0004's forgiving parse. **Tightening it turns five tests red**, including one that exists solely to fail if someone does.

**The serial rule is a principle, not a number** — [serial-disclosure.ts](packages/db/src/domain/serial-disclosure.ts). The plan said "last four when ≥ 8" *and* "revealing half is not masking". At 8, those contradict. Code now reveals the last four only when **more stay hidden than are shown** → first length is 9. Recorded on #32.

**A draft is 404.** A draft claims nothing; publishing it publishes a claim nobody made.

**`registeredOn` is a date, not a timestamp.** ADR 0004 kept issuance time out of the identifier; a public millisecond would put it back.

**The claims have no aggregate verdict.** The serial is `recorded`, never `verified` — the issuer supplied it, nothing checked it against the object. That one word is the difference between this and a green checkmark.

## Structural guarantees, not review habits

- **Internal keys never leave Postgres.** The repository names its columns; `id` and `issuerId` are not in the row. Adding `id` to the projection fails a test.
- **The whole serial is not a field** on `PublicPassport`. A route cannot return what the type does not contain.
- **An unparsed string cannot reach the database.** The lookup takes the branded `TrustPassId`; a test asserts the dependency is never called for a bad identifier.

## Tests

229 pass (169 db, 60 api). New: 65.

Every critical guard was **mutation-checked** — broken on purpose to prove the test can fail:

| Mutation | Result |
| --- | --- |
| remove the serial-disclosure guard | 6 red |
| report the serial as `verified` | 5 red |
| add `id` to the repository projection | 1 red |
| tighten the path-param schema | 5 red |

## Security impact

The first public read surface, so this is where exposure is decided.

- Internal keys and the whole serial are structurally excluded, and asserted
- Every response is `no-store`, the 404 included: a cached "active" after a theft report is the harm this prevents
- Error messages never echo input; a public page renders them
- CORS is open and the endpoint is unauthenticated **by design** — the only protection for a passport is its 128-bit identifier (ADR 0004). No rate limiting yet (TP-103). **The API still must not be exposed publicly before TP-141.**

## Breaking changes

None. New endpoint.

---

- [ ] CI is green
- [x] `pnpm lint`, `pnpm typecheck`, `pnpm test` and `pnpm build` pass locally, from a virgin volume


apps/api/src/routes/passports.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Deliberately permissive. Bounding the length is all this does.
+ *
+ * A format regex here would send every bad identifier through the app's
+ * `defaultHook` and out as `validation_failed`, erasing the one distinction this
+ * endpoint exists to make: a mistyped code is not an unregistered product.
+ * `parseTrustPassId` in the handler decides.
+ */
+const PassportParamsSchema = z.object({
```
clevervi, line 14: **The line most likely to be "fixed" into a bug.** It looks under-validated. A format regex here sends bad IDs through `defaultHook` as `validation_failed` — erasing typo vs. unknown — and rejects lower case. Tightening it turns five tests red, one of which exists only to catch this.

packages/db/src/domain/serial-disclosure.ts
```diff
@@ -0,0 +1,55 @@
...
+  | { readonly disclosed: false }
+  | {
+      readonly disclosed: true;
+      readonly suffix: string;
+      /**
+       * Published deliberately. Rendering four dots for a twelve-character serial
+       * is a small misstatement on a page whose whole job is not to misstate, and
+       * a serial's length is printed on every box of that model anyway.
+       */
+      readonly hiddenCharacters: number;
```
clevervi, line 22: **Publishes the serial's length.** I think that's right: showing four dots for a 31-character serial misstates on a page about not misstating, and length is uniform per model. But it is information the old plan didn't give out. Easy to drop if you disagree.

apps/api/src/passports/read-passport.ts
```diff
@@ -0,0 +1,90 @@
...
+ * would also confirm the identifier is allocated, which a stranger has no
+ * business learning.
+ */
+export async function readPassport(
+  db: Database,
+  trustpassId: TrustPassId,
+): Promise<ReadPassportResult> {
+  const record = await findProductByTrustPassId(db, trustpassId);
+
+  if (!record || record.status === "draft") {
```
clevervi, line 66: **Draft → 404, same code as "never registered".** Deliberate: a distinct code would confirm the ID is allocated. The cost is an issuer can't tell a draft from a typo'd lookup by this endpoint alone — they have their own registration response for that.

apps/api/src/passports/read-passport.ts
```diff
@@ -0,0 +1,90 @@
...
+    passport: {
+      trustpassId: record.trustpassId,
+      brand: record.brand,
+      model: record.model,
+      category: record.category,
+      status: record.status,
+      serial: disclosure.disclosed
+        ? { suffix: disclosure.suffix, hiddenCharacters: disclosure.hiddenCharacters }
+        : null,
+      registeredOn: record.createdAt.toISOString().slice(0, 10),
```
clevervi, line 83: **UTC date.** A product registered at 21:00 in Bogotá shows the next day. Honest about the instant, slightly surprising to a local reader. Alternative is the issuer's country timezone — more machinery than it seemed worth now.

clevervi (commented):
Second pass before handing over. Four calls inline worth disagreeing with. Nothing broken that I could find; every guard below was broken on purpose to confirm its test fails.

I wrote this diff, so this is the same judgement twice, not a second opinion.

Issue #38, opened 2026-09-17
[PASSPORT] The 404 passport page renders nothing without JavaScript


clevervi:
## What happens

`GET /trustpass/<valid but unissued TrustPass ID>` answers **404**, which is
correct. But the HTML body is an empty Suspense placeholder:

```html
<div hidden=""><!--$--><!--/$--></div>
```

The copy in `apps/web/app/trustpass/[trustpassId]/not-found.tsx` is present only
in the RSC Flight payload and is rendered by client JavaScript. A reader with
JavaScript disabled sees a blank page where this project promises a sentence.

## What was measured

Against a **production** build (`next build` + `next start`), not the dev
server, and not assumed:

| Configuration | Status | `<h1>` in HTML |
| --- | --- | --- |
| `notFound()` in the page body, boundary in the dynamic segment | 404 | none |
| `notFound()` in `generateMetadata` | 404 | none |
| Boundary moved to `app/not-found.tsx` | 404 | none |
| Without `export const dynamic = "force-dynamic"` | 404 | none |
| Normal return (typo page, unreadable page) | 200 | rendered |
| Next's own `/_not-found` for an unmatched route | 404 | rendered |

So it is not the placement of the boundary, not `force-dynamic`, and not this
route's code: in Next 16.3.5, `notFound()` thrown from a dynamic App Router page
does not server-render its boundary.

## Why it is not blocking v0.3.0

- The status code is correct, which is what a monitor, a crawler and `curl` read.
- Passports are `noindex`, so no search engine is reading this page anyway.
- The reader this milestone is built for is scanning a QR with a phone, and has
  JavaScript.

## Why it should not be left alone

A blank page is ambiguous, and this project's whole argument is that ambiguity
about a product must never be left for the reader to fill in. "No passport for
this code" and an empty page are different messages.

## Not the fix

Returning the not-found UI as a normal render would make the HTML correct and
the status **200**. A resource that genuinely does not exist answering 200 is a
worse lie than a blank page, and it would break any future monitoring.

## Next step

Reproduce against a newer Next release and, if it persists, open an upstream
issue with the table above.

clevervi:
Removed from the v0.3.0 milestone and left open deliberately.

v0.3.0 shipped with this recorded as a known limitation in both `CHANGELOG.md` and the release notes, rather than carried quietly. It is not assigned to a milestone because the fix depends on an upstream Next release, and putting it in v0.4.0 would be a commitment to a date nobody here controls.

[...]

PR #39, opened 2026-09-17
feat: render the public passport, and say what has not been checked

### LINKED ISSUE #31: [PASSPORT] TP-030 Resolve a TrustPass ID to a public passport

### LINKED ISSUE #32: [PASSPORT] TP-031 Show the product identity, with the serial masked

### LINKED ISSUE #33: [PASSPORT] TP-033 State each verification claim separately

clevervi:
Closes #31. Closes #32. Closes #33.

#36 built the read path. This puts a page on it: the first thing in this
repository a stranger can look at and understand.

## The shape of it

`apps/web/app/trustpass/` — three modules and a page, no new dependency.

- **`passport.ts`** — fetches from the API and **never throws**. It returns a
  discriminated union because the page must distinguish four answers, and a
  thrown error collapses them into one.
- **`claim-wording.ts`** — the only place a code becomes language. ADR 0003 puts
  codes in the API and words in the page so copy can change without a contract
  change.
- **`[trustpassId]/page.tsx`** — a server component rendering the four states.
- **`[trustpassId]/not-found.tsx`** — for a well-formed code that was never
  issued.

## The decision this PR is really about

A mistyped code and an unissued code are **different answers**, and the page
says so:

| Input | Status | What the reader is told |
| --- | --- | --- |
| A real passport | 200 | The passport |
| One character wrong | 200 | "This code has a typo — **this does not mean the product is unregistered**" |
| Valid, never issued | 404 | "No passport for this code" |
| Not a TrustPass code | 200 | "This isn't a TrustPass code" |
| TrustPass unreachable | 200 | "**No check has been made**, so nothing about this product — good or bad — should be read from this page" |

A 404 for a mistyped code would be the system agreeing that the product is
unregistered because someone fat-fingered one character. That false accusation
is the entire reason ADR 0004 spends a check symbol on the identifier, and this
is the first place it pays off.

The unreachable case matters just as much. A passport-shaped layout with blank
fields when the API is down looks like a product with nothing verified about it.

## Deliberately not done

**No `app/robots.ts` disallowing `/trustpass/`**, though the plan called for one.
A crawler blocked by robots.txt never fetches the page, never sees the `noindex`
directive, and can still list the URL from an inbound link. The two cancel each
other. `noindex` is on every branch of the page instead.

## Mutation checks

Every guard here was broken on purpose to prove its test can fail.

| Mutation | Result |
| --- | --- |
| Remove the serial-disclosure guard | 6 tests red |
| Report the serial claim as `verified` | 5 tests red |
| Add `id` to the repository projection | 1 test red |
| Tighten the route's param schema to a format regex | 5 tests red |
| Treat any 404 as `not_found` in the fetch module | 2 tests red |
| Word "Recorded, not checked" as "Verified by issuer" | 2 tests red |

## Verified against a running stack, not just a build

A production build (`next build` + `next start`) served against the real API and
a seeded database — a verified issuer and an unverified one:

```
--- verified issuer -> HTTP 200
    ok  brand and model shown          ok  whole serial absent
    ok  serial masked to last four     ok  internal key absent
    ok  characters withheld stated     ok  no authenticity badge
    ok  registration number shown      ok  claim: recorded, not checked
    ok  country resolved to a name     ok  claim: not independently verified
    ok  noindex
--- unverified issuer -> HTTP 200     banner + "the issuer's own word" both shown
--- one character wrong -> HTTP 200   "This code has a typo"
--- valid but unknown -> HTTP 404     API: passport_not_found
--- "hello" -> HTTP 200               "This isn't a TrustPass code"
```

Two things that looked like defects were my test script being wrong, and are
worth recording so nobody re-finds them:

- The mask did not match `••••4896` because React separates adjacent SSR text
  nodes with `<!-- -->`. The page was correct.
- My "valid but unknown" identifier had a **wrong check symbol**, so the system
  correctly answered `mistyped`. The 26-zero body decodes to `0`, whose check
  symbol is `0`, not `C`. The check symbol caught my typo in a test about
  catching typos.

## One real finding, not fixed here

**#38** — the 404 page answers 404 correctly but renders no HTML; Next 16.3.5
leaves the not-found boundary for the client. Measured across four
configurations against a production build. Not blocking: the status is right,
passports are `noindex`, and a QR reader has JavaScript. Not silently accepted
either — it is written in the code where the decision lives.

## Checklist

- [x] `pnpm test` — 268 tests (39 web, 169 db, 60 api)
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] No lockfile change (no new dependency)
- [x] Verified against a running production build, not only a compile


apps/web/app/trustpass/passport.ts
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * The outcome is decided by the API's error **code**, never by the HTTP status
+ * alone and never by the message. The distinction that matters most: a failure to
+ * reach or understand the API must come back as `unavailable`, never as
+ * `not_found`. "We could not check" and "this product is not registered" are
+ * different answers, and giving the second when the truth is the first tells a
+ * buyer a genuine product is fake.
+ *
+ * That is also why a generic 404 — an API without this route, say — is
+ * `unavailable`. Only the specific `passport_not_found` code means not found.
```
clevervi, line 104: This is the line the milestone turns on. A generic 404 — the API answering `not_found` because the route is missing — must not reach the page as `not_found`, or an outage reads as "this product is not registered".

apps/web/app/trustpass/claim-wording.ts
```diff
@@ -0,0 +1,110 @@
...
+  "serial:recorded":
+    "The issuer supplied this serial. Nothing has compared it to the object in front of you — compare the last characters with the label yourself.",
+  "secure_tag:not_present":
+    "This product has no cryptographic tag. A QR code can be copied onto any object, so it proves only that someone had the code.",
+  "warranty:not_recorded": "No warranty has been recorded for this product.",
+  "physical_authenticity:not_verifiable":
+    "TrustPass has not inspected this object and cannot. It records who made each claim about a product, not whether the object is genuine.",
+};
+
+const UNKNOWN_STATE = { label: "Not described", tone: "neutral" } as const;
```
clevervi, line 56: Unknown state falls back to neutral, never positive. If the API ships a claim state before this page learns it, the worst outcome is "Not described" — not a green tick over something nobody checked.

apps/web/app/trustpass/[trustpassId]/page.tsx
```diff
@@ -0,0 +1,369 @@
...
+  // notFound() works by throwing, so it must not sit inside a try. The fetch
+  // module returns a union instead of throwing for exactly that reason.
+  //
+  // Known limitation, measured against a production build rather than assumed:
+  // Next 16.3.5 answers 404 but does not server-render the not-found boundary
+  // into the HTML — the body arrives as an empty Suspense placeholder and the
+  // copy is rendered on the client from the Flight payload. Verified to be
+  // independent of `force-dynamic` and of whether the boundary sits in this
+  // segment or at the app root. A reader without JavaScript therefore sees a
+  // blank 404 rather than the sentence in `not-found.tsx`. Tracked as #38.
```
Comment (clevervi) [inline apps/web/app/trustpass/[trustpassId]/page.tsx:351]: Measured, not assumed: four configurations against a production build, all 404 with an empty body. Left in the code rather than only in #38 because this is where someone would otherwise "fix" it by returning 200.

PR #40, opened 2026-09-17
feat: put a scannable code on the passport, and say what a scan proves

### LINKED ISSUE #34: [PASSPORT] TP-032 Generate the QR that resolves to the passport
## Objective

Generate a QR code that resolves to a product's passport.

## Scope

**In scope:**

- A pure function encoding a passport URL as SVG
- The QR rendered inline on the passport page
- A route handler serving the same SVG as a file, for issuers who need one to print

**Out of scope:**

- Printing, labels, or bulk generation
- Secure NFC (TP-130+), which is a different assurance tier entirely

## Context

**This is a web task, not an API one, and the reason is not preference.**

Hono's `secureHeaders()` sets `Cross-Origin-Resource-Policy: same-origin` by default. Verified in `hono/dist/middleware/secure-headers/secure-headers.js`:

```js
crossOriginResourcePolicy: ["Cross-Origin-Resource-Policy", "same-origin"],
// DEFAULT_OPTIONS: crossOriginResourcePolicy: true
```

A page served from `:3000` embedding `` is blocked by the browser. CORS does not help — CORP is a separate mechanism.

Putting the QR in the web app is also the better design on its own merits: a QR is an encoding of a URL, and the web app owns the URL. Rendering it inline server-side means no second request, no client JavaScript, and no cross-origin problem.

**Dependency choice.** `qrcode` — the popular one — pulls in `pngjs`, `dijkstrajs` and `yargs`, a command-line argument parser, none of which belong in a server bundle. `qrcode-generator` has **zero** dependencies and emits SVG directly.

## Acceptance criteria

- [ ] A pure function takes a URL and returns well-formed SVG
- [ ] The function is unit-tested, including that it encodes the URL it was given
- [ ] The passport page renders the QR inline, server-side, with no client JavaScript
- [ ] A route handler serves the same SVG with the correct content type
- [ ] The QR encodes the absolute passport URL, not a relative path
- [ ] Exactly one new dependency is added, and it has no transitive dependencies
- [ ] **Scanning the rendered code with a phone lands on the passport page**

## Dependencies

TP-030

## Component

web

## Risk

low

## Estimate

3

clevervi:
Closes #34.

The last piece of v0.3.0. A product now has something you can point a camera at.

## Where it lives, and why it is not in the API

Hono's `secureHeaders()` sets `Cross-Origin-Resource-Policy: same-origin`. A page
on `:3000` embedding `` is blocked by
the browser, and CORS does not help — CORP is a separate mechanism.

It is also the better design on its own merits: a QR is an encoding of a URL,
and the web app owns the URL. The API never learns what the web's routes look
like.

## One dependency, verified rather than trusted

`qrcode-generator@2.0.4`, checked against its manifest rather than its README:

```
dependencies: {}    peerDependencies: {}
```

The popular `qrcode` pulls in `pngjs`, `dijkstrajs` and `yargs` — a command-line
argument parser — into a server bundle.

`jsqr@1.4.0` is added as a **dev** dependency, also dependency-free, and is the
reason the tests are worth anything. See below.

## The bug that was waiting to happen

`qrcode-generator`'s default byte encoder keeps the low byte of each code unit
and discards the rest. Measured, not inferred:

```js
qrcode.stringToBytes("café 中")  // → [99, 97, 102, 233, 32, 45]
```

`中` (U+4E2D) became **45** — an ASCII hyphen. No error. The resulting code
scans perfectly, to a **different URL**. A QR printed on a product that resolves
somewhere else is worse than no QR at all, so `encodeQr` refuses anything
outside printable ASCII instead of doing its best.

A passport URL is ASCII by construction, so the only way in is a misconfigured
base URL — which is exactly when you want a loud failure.

## Tested by decoding, not by re-encoding

Reading a QR back with the library that wrote it proves only that the library
agrees with itself. These tests rebuild the module grid one pixel per module and
hand it to **jsQR**, an unrelated decoder, asserting it reads back the exact URL
given. No rasterisation needed — jsQR takes raw RGBA.

## Mutation checks

| Mutation | Result |
| --- | --- |
| Remove the printable-ASCII guard | 4 tests red |
| `QUIET_ZONE` 4 → 0 | 1 test red |
| `QUIET_ZONE` 4 → 2 | 1 test red |
| Make the QR follow the theme (`currentColor`) | 1 test red |
| Remove XML escaping from the description | 1 test red |
| Remove trailing-slash handling in `passportUrl` | 1 test red |

**The quiet-zone test failed to fail on the first attempt, and that is the
finding.** It asserted the margin against `code.quietZone` — the value under
test — so setting the margin to zero made the assertion trivially true. It now
asserts against the literal specification minimum of 4 and catches both removing
the margin and halving it. This is the entire argument for breaking your own
guards on purpose: the test looked thorough and checked nothing.

```
--- the downloadable file
    ok  HTTP 200                    ok  black on white
    ok  served as SVG               ok  decodes to this passport's URL
    ok  cached immutably, unlike the passport itself
--- the QR rendered inline on the page
    ok  an inline QR is present     ok  hardcoded black, not theme-following
    ok  decodes to the same URL     ok  a download link is offered
    ok  rendered as SVG elements, not injected HTML
    ok  the page states a scan proves only that someone had the code
--- a rubbish identifier
    ok  over-long identifier -> HTTP 404
```

Both surfaces were decoded by jsQR from the bytes the server actually sent.

**Not claimed: that it scans on a physical phone.** Decoding the matrix proves
the encoding is correct; it does not prove scanner compatibility across camera
apps and print sizes. That is a hardware check and it has not been done.

## Two smaller decisions

**The route does no lookup.** A QR is an encoding of a URL, and the page that
URL points to already says honestly whether a passport exists. Checking here
would make printing a label depend on the API being reachable, and would answer
a question the route was not asked. The identifier is length-capped instead,
because a QR grows with its input.

**Immutable caching on the file, `no-store` on the passport.** Trust data
changes — a suspended product must read as suspended immediately. This image is
a pure function of the identifier in its own URL and can never change for it.

## Checklist

- [x] `pnpm test` — 283 tests (54 web, 169 db, 60 api)
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Lockfile committed with the dependency change
- [x] Verified against a running stack, not only a compile


apps/web/lib/qr.ts
```diff
@@ -0,0 +1,134 @@
...
+ * of each code unit and discards the rest. `"中"` (U+4E2D) becomes byte 0x2D,
+ * which is `-`. No error is raised, and the resulting code scans perfectly —
+ * to a different URL than the one requested.
+ *
+ * A passport URL is ASCII by construction (a host plus a Crockford base32
+ * identifier), so the only way to get here is a misconfigured base URL. Failing
+ * loudly is the whole point: a QR on a product that resolves somewhere else is
+ * worse than no QR.
+ */
+const PRINTABLE_ASCII = /^[\x20-\x7e]+$/;
```
clevervi, line 62: This guard is the whole reason this module exists rather than a one-line call to the library. Silent corruption into a valid-but-wrong URL is the worst possible failure for a code printed on a physical object.

apps/web/lib/qr.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    const coordinates = [...code.path.matchAll(/M(\d+) (\d+)h1v1h-1z/g)].map((match) => ({
+      column: Number(match[1]),
+      row: Number(match[2]),
+    }));
+
+    // Four is the specification's minimum, asserted as the literal number
+    // rather than against `code.quietZone`. Deriving the bound from the value
+    // under test makes the assertion true for any margin including none, which
+    // is exactly what an earlier version of this test did.
+    const SPECIFIED_MINIMUM = 4;
```
clevervi, line 62: Asserted against the literal 4, not against `code.quietZone`. The earlier version derived the bound from the value under test, so it passed with no margin at all — the mutation check caught it, the test suite did not.

apps/web/app/trustpass/[trustpassId]/qr.svg/route.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return new Response(svg, {
+    headers: {
+      "content-type": "image/svg+xml; charset=utf-8",
+      // The opposite of the passport page's `no-store`, and deliberately so.
+      // The passport is never cached because trust data changes — a suspended
+      // product must read as suspended immediately. This image is a pure
+      // function of the identifier in the URL and can never change for it.
+      "cache-control": "public, max-age=31536000, immutable",
```
Comment (clevervi) [inline apps/web/app/trustpass/[trustpassId]/qr.svg/route.ts:45]: Deliberate asymmetry with the passport page, which is `no-store`. Worth stating in code so nobody "fixes" the inconsistency later: the passport carries trust data that changes, this image cannot.

PR #44, opened 2026-09-17
fix: a QR must never point at a host this deployment does not own

### LINKED ISSUE #43: [PASSPORT] A QR built without NEXT_PUBLIC_SITE_URL silently encodes the wrong host
## Objective

A passport QR must never resolve to a host that is not this deployment. Today it
can, silently, and the failure is invisible until somebody scans the code.

## The defect

`apps/web/lib/qr.ts`:

```ts
export function passportUrl(trustpassId: string, baseUrl?: string): string {
  const base = baseUrl ?? process.env.NEXT_PUBLIC_SITE_URL ?? "http://localhost:3000";
  ...
}
```

Both call sites — the inline QR on the passport page and the `qr.svg` route —
call it with no `baseUrl`. When `NEXT_PUBLIC_SITE_URL` is unset, every QR
encodes `http://localhost:3000` regardless of where the app is actually served.

**It scans cleanly.** Nothing errors, nothing logs, and the code is
indistinguishable from a correct one until a camera follows it somewhere else.

This is the same class of defect as the encoder truncation guarded in #40, where
`中` became an ASCII hyphen and produced a valid code for a different URL. That
one was guarded; this one was left, in the same file, in the same release.

### Why it is not merely theoretical

`http://localhost:3000` is not a dead address on a developer's machine — it is
whatever else is bound to that port. During v0.3.0 a second project on this
machine defaulted to 3000, so a QR generated with the variable unset would have
resolved to an unrelated application's home page.

## Scope

**In scope:** how the base URL is decided, at both call sites, and what the
`qr.svg` route is allowed to cache as a result.

**Out of scope:** the QR encoding itself, which is correct and tested.

## The constraint that makes this non-trivial

The obvious fix — derive the origin from the request's `Host` header — is unsafe
as written, because `qr.svg` sets:

```
cache-control: public, max-age=31536000, immutable
```

A request carrying a forged `Host` would be encoded into the SVG and then cached
under that URL for a year. Every later visitor would be served a QR pointing at
the attacker's host. That converts a misconfiguration into cache poisoning, so
the two decisions — where the origin comes from, and what may be cached — have
to be made together.

## Acceptance criteria

- [ ] `passportUrl` takes the base URL as a **required** argument and reads no
      environment variable. A pure function must not guess.
- [ ] When `NEXT_PUBLIC_SITE_URL` is set, it is used, and nothing about the
      request can override it.
- [ ] When it is unset **in development**, the origin is derived from the
      request so the QR works on any port.
- [ ] When it is unset **in production**, no QR is rendered at all. The passport
      page still renders everything else; the `qr.svg` route answers an error.
      A missing QR is honest. A QR to the wrong host is not.
- [ ] A response whose origin came from the request is **never** cached
      immutably. Only a configured origin may be.
- [ ] Tests cover each branch, including that a forged `Host` cannot reach a
      cacheable response.
- [ ] Every guard mutation-checked, with the table in the pull request.

## Risk

`risk:high`. This decides what a physical, printed, permanent object resolves
to. A wrong QR cannot be recalled once a label is on a product.

clevervi:
Closes #43.

## The defect

`passportUrl` defaulted to `http://localhost:3000` when `NEXT_PUBLIC_SITE_URL`
was unset, and **both** call sites called it with no base URL. Every QR encoded
that host wherever the app was actually served.

It scanned cleanly. Nothing errored, nothing logged, and the code was
indistinguishable from a correct one until a camera followed it.

Same class as the encoder truncation guarded in #40 — `中` became an ASCII
hyphen and produced a valid code for a different URL. **I guarded that one and
left this one, in the same file, in the same release.**

Not theoretical: `localhost:3000` is not a dead address on a developer's
machine, it is whatever else holds the port. During v0.3.0 a second project on
this machine defaulted to 3000.

## Why the obvious fix is unsafe

Deriving the origin from the request's `Host` is right for whoever is looking at
the page — but `qr.svg` sets `public, max-age=31536000, immutable`, and `Host`
is caller-controlled. A forged `Host` would be encoded into the SVG and cached
under that URL for a year, serving every later visitor a code pointing at the
attacker. That converts a misconfiguration into cache poisoning.

So the two decisions had to be made together: **where the origin comes from**
determines **what may be cached**.

## The shape

`lib/passport-origin.ts` returns a tagged union, and the tag is the point:

| Origin | Source | `Cache-Control` |
| --- | --- | --- |
| `configured` | `NEXT_PUBLIC_SITE_URL` | `public, max-age=31536000, immutable` |
| `request` | this request's `Host`, **development only** | `no-store` |
| `unknown` | nothing usable | `no-store`, and no QR is rendered |

`passportUrl` is now pure with a **required** base URL and reads no environment
variable. Deciding the origin is a question about the request and the
deployment; a pure function can see neither, so it must not guess.

With no origin, the page renders the passport **without** the QR and the route
answers **503**, not 404 — the passport exists, this deployment just cannot say
where it lives, and a 404 invites the caller to cache that belief. A missing QR
is honest. A QR to the wrong host is a lie printed on an object that cannot be
recalled.

## Mutation checks

| Mutation | Result |
| --- | --- |
| Production falls back to the request `Host` | 1 test red |
| Cache a request-derived origin immutably | 2 tests red |
| Remove `Host` validation | 5 tests red |
| Remove the scheme check on the configured URL | 2 tests red |
| Remove the trailing-slash trim | 1 test red |

## Verified against a production build, with a genuinely forged Host

**`fetch` cannot forge `Host` — it is a forbidden header name, and it silently
ignores it.** My first attempt proved nothing. These used a raw socket.

**Production, `NEXT_PUBLIC_SITE_URL=https://trustpass.example`:**

```
Host: evil.example                    -> 200, immutable
  decodes to https://trustpass.example/trustpass/TP1-7TB9...
Host: trustpass.example.attacker.net  -> 200, immutable
  decodes to https://trustpass.example/trustpass/TP1-7TB9...
```

**Production, variable unset:**

```
page   -> 200, passport renders in full, QR absent, no "localhost:3000" anywhere
qr.svg -> 503, cache-control: no-store
```

**Development on port 3002, variable unset — the original bug:**

```
qr.svg -> 200, cache-control: no-store
  decodes to http://localhost:3002/trustpass/TP1-7TB9...   (was localhost:3000)
```

**Development, forged Host, variable unset:**

```
Host: evil.example -> 200, cache-control: no-store
  decodes to http://evil.example/trustpass/TP1-7TB9...
```

That last one is worth stating plainly rather than hiding: in development
without configuration, a forged `Host` **does** reach the code. It is served
`no-store`, so it cannot be stored and replayed to anyone else — the attacker
poisons only their own response. The vector is closed by the caching rule, not
by rejecting the host, and that is the design rather than an oversight.

## Checklist

- [x] `pnpm test` — 305 tests (76 web, 169 db, 60 api)
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Verified against a production build and a real database, not only a compile


apps/web/lib/passport-origin.ts
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * So both are used, and the resulting origin says which it was, because the
+ * caller must cache a request-derived answer differently from a configured one.
+ */
+
+export type PassportOrigin =
+  /** From `NEXT_PUBLIC_SITE_URL`. Independent of the request, so safe to cache. */
+  | { readonly kind: "configured"; readonly baseUrl: string }
+  /** Derived from this request's `Host`. Never cacheable, and never in production. */
+  | { readonly kind: "request"; readonly baseUrl: string }
```
clevervi, line 25: The tag is the whole design. Without it the caller cannot tell a configured origin from a request-derived one, and would have to cache both the same way — which is how the poisoning vector gets reopened by someone tidying this up later.

apps/web/app/trustpass/[trustpassId]/qr.svg/route.ts
```diff
@@ -24,25 +25,47 @@ export async function GET(
...
   }
 
+  const origin = resolvePassportOrigin({ host: request.headers.get("host") });
+
+  // 503 rather than 404: the passport exists, this deployment just cannot say
+  // where it lives. A 404 would claim there is nothing here, and the caller
+  // would cache that belief.
+  if (origin.kind === "unknown") {
+    return new Response("Passport origin is not configured; set NEXT_PUBLIC_SITE_URL.", {
+      status: 503,
```
Comment (clevervi) [inline apps/web/app/trustpass/[trustpassId]/qr.svg/route.ts:35]: 503 and not 404 on purpose. A 404 asserts nothing is here, the caller may cache that, and the passport does in fact exist — this deployment just cannot say where it lives.

apps/web/app/trustpass/[trustpassId]/page.tsx
```diff
@@ -292,10 +294,19 @@ function Passport({ passport }: { passport: PassportView }) {
...
  * something already computed here.
  */
-function PassportQr({ trustpassId }: { trustpassId: string }) {
+async function PassportQr({ trustpassId }: { trustpassId: string }) {
+  const origin = resolvePassportOrigin({ host: (await headers()).get("host") });
+
+  // No origin, no QR — and the rest of the passport is unaffected. Everything
+  // else on this page is still true; a code pointing at the wrong host would
+  // not be, and it would be the one part a reader acts on.
+  if (origin.kind === "unknown") {
```
Comment (clevervi) [inline apps/web/app/trustpass/[trustpassId]/page.tsx:303]: Returning null here rather than throwing: a wrong QR is the one element on this page a reader acts on, so it is the one worth dropping. Everything else on the passport is still true without it.

PR #52, opened 2026-09-17
feat: record what happened to a product, and why

### LINKED ISSUE #51: [LIFECYCLE] TP-050 Record what happened to a product, and why
## Objective

Give the system an append-only record of what happened to a product — including
**why** — so that every milestone after this one writes to one spine instead of
inventing a parallel history.

## Why now

[`product-lifecycle.md`](../blob/develop/docs/product-lifecycle.md) states that a
status change leaves no record of why it happened. That is not a future gap:
products can be suspended **today**, and the passport can render `Suspended`
while being unable to say whether that was a theft report, a fraud investigation
or a disputed claim. To a buyer those are very different facts.

## Scope

**In scope:** the `lifecycle_event` table, its closed sets, the append-only
guarantee, and the domain types.

**Out of scope:**

- Writing events from the registration path — the table first, wired up next
- Rendering history on the passport (`TP-034`)
- Who may record what (`TP-053`)
- Ownership events (`TP-060`+)

## The shape, per [ADR 0008](../blob/develop/docs/adr/0008-events-record-what-happened-claims-assert-what-is-true.md)

| Field | Note |
| --- | --- |
| `product` | restrict, never cascade — deleting provenance is the one thing this exists to prevent |
| `type` | closed set |
| `actor_kind` | what capacity the actor acted in. Identity of a *person* waits for `TP-141` |
| `issuer_id` | nullable, set when the actor was an issuer |
| `occurred_at` | when it happened in the world |
| `recorded_at` | when TrustPass learned of it |
| `reason` | closed set, nullable |
| `source_reference` | what backs it, where that exists |
| `previous_state` / `resulting_state` | nullable — not every event is a transition |

## Acceptance criteria

- [ ] `lifecycle_event` exists with a `bigint identity` primary key that never
      crosses the API boundary (ADR 0005)
- [ ] `occurred_at` and `recorded_at` are **separate columns**, both `NOT NULL`
- [ ] A database check enforces `occurred_at <= recorded_at`. An event is
      past-tense by ADR 0008; a future occurrence is not an event
- [ ] `type`, `reason` and `actor_kind` are Postgres enums, not free text
- [ ] **UPDATE and DELETE are refused by a trigger**, with a project SQLSTATE,
      not by convention. Append-only enforced by application discipline is
      enforced by nothing
- [ ] A correcting event is the way to fix a mistake, and a test demonstrates it
- [ ] `product_id` is `RESTRICT`
- [ ] The migration is named, forward-only, and applied against real Postgres in CI
- [ ] Integration tests run against real Postgres and skip cleanly without it

## Deliberately narrow enum sets

Postgres adds an enum value cheaply and cannot remove one, so the initial sets
cover what exists today and nothing speculative — the same reasoning
`product_category` already records.

## Risk

`risk:high`. The history is the product. After rows exist the shape is migrated,
not edited, and by construction those rows are not supposed to be rewritten.

clevervi:
Closes #51.

The event spine. [ADR 0008](docs/adr/0008-events-record-what-happened-claims-assert-what-is-true.md)
decided the shape; this builds it, before ownership, warranty, repair or
disputes have a chance to invent their own.

## The gap this closes

[`product-lifecycle.md`](docs/product-lifecycle.md) already stated that a status
change leaves no record of why it happened. That was not a future problem:
products can be suspended **today**, and the passport can render `Suspended`
while being unable to say whether that was a theft report, a fraud investigation
or a disputed claim. To a buyer those are very different facts, and the worst
reading was the one available.

## Append-only, enforced where it cannot be bypassed

A rule that lives only in TypeScript is bypassed by the first path that writes
without going through it. Here that matters more than it does for a status,
because **an editable history is not a history** — it is a claim about the past
that anyone with a connection can rewrite.

```
UPDATE  -> TP002  "cannot be modified. Record a correcting event instead."
DELETE  -> TP002  "cannot be deleted. History is append-only."
TRUNCATE-> TP002
```

Three things worth defending:

- **Not `REVOKE`.** A grant is scoped to a role and the application connects as
  the owner, which can re-grant itself anything. A `BEFORE` trigger fires for
  every writer, owner included.
- **`TRUNCATE` gets its own statement-level trigger.** It is not a `DELETE` and a
  row-level trigger never sees it. Without it the entire history is one
  statement away from gone.
- **No column list on the `UPDATE` trigger.** A guard covering `reason` but not
  `occurred_at` protects nothing.

`TP002` is deliberately distinct from the `TP001` the status guard raises, so a
caller can tell "that move is not allowed" from "history cannot be rewritten"
without parsing a message.

## Two decisions that will feel like ceremony later

**`occurred_at` and `recorded_at` are separate columns.** A repair done in March
and recorded in September is two dates; storing one forces a choice between
lying about when it happened and lying about when it became known. The distance
between them is itself information — a sale recorded two years late is a weaker
record than one recorded the same day.

**`reason` is a closed set.** The first time someone writes "suspended pending
investigation into possible theft", the system has a paragraph where it needed
`theft_report`. Reasons are also **not states**: `suspended` stays one status,
and what distinguishes a theft report from a warranty dispute lives in history,
the one place that does not get overwritten.

## Mutation checks

Every guard broken on purpose against real Postgres, each restored after.

| Mutation | Result |
| --- | --- |
| Drop the append-only `UPDATE` trigger | 3 tests red |
| Drop the append-only `DELETE` trigger | 1 test red |
| Drop `lifecycle_event_not_in_future` | 1 test red |
| Drop `lifecycle_event_issuer_matches_actor` | 3 tests red |
| Drop `lifecycle_event_correction_targets` | 5 tests red |
| Drop `lifecycle_event_transition_complete` | 7 tests red |
| Product FK `RESTRICT` → `CASCADE` | 8 tests red |

That last one is the one I most wanted to see fail: under `CASCADE`, deleting a
product silently deletes its history, which is the single thing a passport
exists to preserve.

## Verified against real Postgres, from a virgin volume

`pnpm db:reset` then the suite — which also proves the hand-written `0005`
migration applies cleanly from zero alongside the generated `0004`.

```
packages/db   192 passed        apps/api   60 passed
apps/web       76 passed        total     328 passed
```

Before the tests existed, all eight guards were probed directly through `psql`
so the constraints were observed firing rather than inferred from a green suite.

## A correction made during the work

My first version of the test called `createDatabase({ url })`. The signature
takes the URL **positionally**, so `postgres()` ignored the object and fell back
to the OS user — the failure read `password authentication failed for user
"darkt"`, which looks like a credentials problem and was a call-shape problem.

Worth recording because the misleading error cost more than the bug.

## Deliberately not here

- Writing events from the registration path — the table first, wired up next
- Rendering history on the passport (`TP-034`)
- Who may record what (`TP-053`)
- Ownership events (`TP-060`+)

The enum sets are narrow on purpose: Postgres adds a value cheaply and cannot
remove one, so they cover what the system can record today and nothing
speculative.

## The test suite does not clean up after itself

Deliberate. Events cannot be deleted, so the product and issuer that own them
cannot be deleted either. **A teardown that succeeded would prove the guarantee
does not hold.**

## Checklist

- [x] `pnpm test` — 328 against Postgres, 305 pass with it absent (skips cleanly)
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Migrations named, forward-only, applied from a virgin volume


packages/db/drizzle/0005_lifecycle_event_append_only.sql
```diff
@@ -0,0 +1,57 @@
...
+EXECUTE FUNCTION lifecycle_event_append_only();
+--> statement-breakpoint
+CREATE TRIGGER lifecycle_event_no_delete
+BEFORE DELETE ON lifecycle_event
+FOR EACH ROW
+EXECUTE FUNCTION lifecycle_event_append_only();
+--> statement-breakpoint
+-- TRUNCATE is not a DELETE and a row-level trigger never sees it. Without this
+-- the whole history is one statement away from gone.
+CREATE TRIGGER lifecycle_event_no_truncate
```
clevervi, line 54: TRUNCATE needs its own statement-level trigger — it is not a DELETE and a row-level trigger never sees it. Without this the entire history is one statement away from gone, and the other two triggers would have looked like they covered it.

packages/db/src/schema/lifecycle-event.ts
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * When this happened in the world.
+     *
+     * Separate from `recordedAt` on purpose. A repair done in March and
+     * recorded in September is two dates, and storing one forces a choice
+     * between lying about when it happened and lying about when it became
+     * known. The distance between them is itself information: a sale recorded
+     * two years late is a weaker record than one recorded the same day.
+     */
+    occurredAt: timestamp("occurred_at", { withTimezone: true }).notNull(),
```
clevervi, line 150: The two timestamps are the decision most likely to be "simplified" by someone later. A repair done in March and recorded in September is two dates, and the distance between them is how you tell a strong record from a weak one.

packages/db/src/schema/lifecycle-event.integration.test.ts
```diff
@@ -0,0 +1,325 @@
+import { desc, eq } from "drizzle-orm";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
```
clevervi, line 2: No teardown, on purpose. If this suite could delete its own rows, the append-only guarantee would not hold — a green teardown would be the bug.

Issue #54, opened 2026-09-17
[LIFECYCLE] Guarantee no product state can change without provenance


clevervi:
## The gap

`TP-051` makes registration write a product and its first event in one
transaction. `TP-052` does the same for a status change. Between them they cover
**every path that exists today**.

Neither stops a future path — a bulk import, a migration, a fixture, a console
session — from creating a product or moving its status with no event. And per
ADR 0008 that failure is invisible: nothing later can distinguish a history that
is *missing* from one that is *empty*.

This is the argument the repository already makes twice, in
`0002_product_status_transition_guard.sql` and
`0005_lifecycle_event_append_only.sql`: a rule that lives only in application
code is bypassed by the first path that writes without going through it.

## The shape

`DEFERRABLE INITIALLY DEFERRED` constraint triggers on `product`, checking at
**commit** that:

- an inserted row has at least one lifecycle event
- a row whose status changed has an event whose `previous_state` and
  `resulting_state` match the move

Deferred because the event is necessarily written after the row it references. A
plain `AFTER` trigger cannot work — the event does not exist yet.

**Check order matters.** The existing transition guard raises `TP001` for an
illegal move. A provenance check that fired first would report "no event" for a
transition that was illegal anyway, masking the thing under test.

## Why this is not part of TP-051 or TP-052

**Blast radius, measured rather than guessed:**

| Suite | Sites | What it is for |
| --- | --- | --- |
| `product.integration.test.ts` | 17 inserts | the product table's own constraints |
| `product-status.integration.test.ts` | 6 status updates | the transition trigger itself |
| others | 6 | assorted fixtures |

The two largest exist to test *one thing each*. Forcing every fixture to also
record provenance would make them assert two things at once and obscure which
one failed.

That is a real cost against a real benefit, and it deserves its own change with
its own diff rather than riding along inside a feature.

## Acceptance criteria

- [ ] A product cannot be committed without at least one lifecycle event
- [ ] A status change cannot be committed without an event recording the move
- [ ] Both constraints are deferred, so the ordinary write order still works
- [ ] The transition guard still reports `TP001` first for an illegal move
- [ ] A project SQLSTATE distinguishes these from `TP001` and `TP002`
- [ ] Existing fixtures updated deliberately, not silenced
- [ ] Mutation-checked: dropping each trigger turns a named test red

## Depends on

`TP-051` and `TP-052`. There is no point guaranteeing provenance before the real
paths produce it.

PR #55, opened 2026-09-17
feat: register a product and where it came from as one act

### LINKED ISSUE #53: [LIFECYCLE] TP-051 Record provenance when a product is registered
## Objective

Make registering a product and recording where it came from **one act**, so the
event spine starts filling instead of standing empty.

## Scope

**In scope:** `insertProduct` writes the product and its first lifecycle event in
one transaction; a reader for a product's history.

**Out of scope:**

- Rendering history on the passport (`TP-034`)
- Events for status changes after creation (`TP-052`)
- Who may record what (`TP-053`)

## Acceptance criteria

- [ ] Registering a product writes **exactly one** event, in the same
      transaction as the product. If the event cannot be written, the product
      must not exist
- [ ] The event names the issuer as actor — an anonymous write records nothing
      worth having
- [ ] The event is dated to the product's own `created_at`, not to a second
      clock read
- [ ] A product created `registered` records `product_registered`; one created
      `draft` records `record_enrolled`, because a draft claims nothing and
      saying it was registered would be a claim nobody made
- [ ] A duplicate live serial still returns `duplicate_live_serial` and leaves
      no product behind — wrapping the writes must not turn a recognised
      rejection into a generic failure
- [ ] The history reader exposes no internal keys (ADR 0005)
- [ ] History is ordered by **when things happened**, not when they were
      recorded
- [ ] A product with no history reads as empty rather than failing

## Risk

`risk:high`. A product with no provenance is invisible: nothing later can tell
that its history is missing rather than simply empty.

clevervi:
Closes #53.

#52 built the event spine and left it empty. This starts filling it: registering
a product and recording its provenance are now **one transaction**. If the event
cannot be written, the product does not exist.

That matters because a product with no provenance is invisible — nothing later
can distinguish a history that is *missing* from one that is *empty*.

## A real defect, found after the suite was green

The first implementation dated the event with the `createdAt` returned from the
insert. It looked right and every test passed. Then I asked Postgres:

```
SELECT e.occurred_at = p.created_at AS event_dated_to_creation ...
-> f
```

**`timestamptz` holds microseconds; a JavaScript `Date` holds milliseconds.**
Round-tripping the value through the client truncated it, so every event landed
up to 999 microseconds *before* the row it described.

My test could not see it: it compared two `Date` objects, both already truncated
to the same wrong value. **It was asserting agreement between two copies of the
same mistake.**

Fixed twice over — the repository now uses `now()`, which is the transaction
timestamp and therefore the exact reading the `created_at` default already took,
and the test now does the comparison in SQL where the precision still exists.

| | Result |
| --- | --- |
| Fixed code | 9 passed |
| Reverted to the JavaScript `Date` | **1 failed**, 8 passed |
| Restored | 9 passed |

That middle row is the point. The old test could not produce it.

## What the transaction has to keep doing

A duplicate live serial still returns `duplicate_live_serial` and leaves no
product behind. Wrapping two writes must not turn a recognised rejection into a
generic failure, which is what happens if the constraint name is lost on the way
out — asserted directly.

## The history reader

Named columns, so a column added to the table later cannot start crossing the
boundary on its own (ADR 0005), and ordered by **when things happened** rather
than when they were recorded. A repair from March learned about in September
belongs in March; ordering by `recorded_at` tells the story of this system's
bookkeeping instead of the story of the product.

## Two API teardowns had to stop deleting

`register-product` and `read-passport` integration tests failed — not the tests,
the `afterAll`:

```
update or delete on table "product" violates RESTRICT setting of
foreign key constraint "lifecycle_event_product_id_product_id_fk"
```

Products now carry history, events cannot be deleted, and the foreign key is
`RESTRICT`. **A teardown that succeeded here would prove a product's history can
be erased**, so the rows stay and each run uses its own prefix. This is the
design asserting itself, not a regression.

## Verified against a running stack, from a reset database

```
POST /products   -> 201  TP1-5J879BCWQXAB5TVF0X7B83W810*
GET  /passports  -> 200  (the passport still reads)
duplicate        -> 409  duplicate_serial, no orphan row

trustpass_id          | type               | actor_kind | actor              | dated_to_creation | past_tense
TP1-5J879BCW...       | product_registered | issuer     | Andes Tech Imports | t                 | t

products with no provenance: 0
```

## Tests

```
packages/db  201 passed     with Postgres:    337 passed
apps/api      60 passed     without Postgres: 205 passed, 132 skipped cleanly
apps/web      76 passed
```

## Deliberately not here

- Events for status changes after creation — `TP-052`
- Rendering history on the passport — `TP-034`
- **A database guarantee that no product can exist without provenance** — raised
  as #54 with the blast radius measured: 23 product-insert call sites, 17 of
  them in a suite whose job is testing the product table's *own* constraints.
  Forcing those fixtures to create events would make them assert two things at
  once. That deserves its own diff.

## Checklist

- [x] `pnpm test` — with and without a database
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Verified over HTTP against a reset database, not only a compile


packages/db/src/repositories/product-repository.ts
```diff
@@ -60,15 +61,45 @@ export async function insertProduct(
...
+        // `now()` rather than the returned `created.createdAt`, and the
+        // difference is not cosmetic. Postgres stores `timestamptz` to
+        // microseconds; a JavaScript `Date` holds milliseconds, so passing the
+        // value back through the client truncates it and the event lands up to
+        // 999 microseconds before the row it describes.
+        //
+        // `now()` is the transaction timestamp and is constant for the whole
+        // transaction, so this is the exact value the `created_at` default
+        // already took — the same clock reading, never a second one.
+        occurredAt: sql`now()`,
```
clevervi, line 98: now() rather than the returned createdAt, and this is the line the PR is really about. A Date truncates to milliseconds and timestamptz keeps microseconds, so the original round-trip put every event slightly before the row it described — and a green suite could not see it.

packages/db/src/repositories/lifecycle-event-repository.integration.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    // Asked of Postgres, not compared in JavaScript.
+    //
+    // An earlier version of this test compared two `Date` objects and passed
+    // while the database disagreed: `timestamptz` holds microseconds, a
+    // JavaScript `Date` holds milliseconds, so both sides had already been
+    // truncated to the same wrong value. The comparison has to happen where the
+    // precision still exists.
+    const [row] = await db.execute<{ aligned: boolean }>(sql`
+      SELECT e.occurred_at = p.created_at AS aligned
```
clevervi, line 102: The comparison has to happen in SQL. Two Dates compared in JavaScript agree with each other while disagreeing with the database, which is how the defect survived a green suite in the first place.

packages/db/src/repositories/lifecycle-event-repository.ts
```diff
@@ -0,0 +1,58 @@
...
+      actorKind: lifecycleEvent.actorKind,
+      occurredAt: lifecycleEvent.occurredAt,
+      recordedAt: lifecycleEvent.recordedAt,
+      reason: lifecycleEvent.reason,
+      previousState: lifecycleEvent.previousState,
+      resultingState: lifecycleEvent.resultingState,
+    })
+    .from(lifecycleEvent)
+    .where(eq(lifecycleEvent.productId, productId))
+    .orderBy(desc(lifecycleEvent.occurredAt), desc(lifecycleEvent.id));
```
clevervi, line 57: Ordered by occurredAt, not recordedAt. A repair from March learned about in September belongs in March — that is where a reader looks for it.

PR #57, opened 2026-09-17
feat: record why a product's status changed

### LINKED ISSUE #56: [LIFECYCLE] TP-052 Record why a product's status changed
## Objective

Make a status change record **who moved it and why**, so the passport can stop
rendering `Suspended` with nothing behind it.

This is the gap [`product-lifecycle.md`](../blob/develop/docs/product-lifecycle.md)
admits: a status change leaves no record of why it happened. A theft report, a
fraud investigation and a disputed warranty claim currently look identical to a
buyer, and they are not remotely the same news.

## Scope

**In scope:** a repository function that changes a product's status and records
the event in one transaction.

**Out of scope:**

- An HTTP endpoint to call it — status changes need authentication (`TP-141`)
  before anything outside can move one
- Rendering the reason on the passport (`TP-034`)
- The database-level guarantee that no status change can happen without an
  event (#54)

## Acceptance criteria

- [ ] Changing a status writes the product update and its event in **one
      transaction**. If the event cannot be written, the status does not move
- [ ] The event carries `previous_state` and `resulting_state`, both set
- [ ] `reason` is required — a status change with no reason is the defect this
      closes, so it cannot be optional here
- [ ] The event type follows **both ends** of the move, not just the
      destination: `draft → registered` is `product_registered`, while
      `suspended → registered` is `product_reinstated`
- [ ] An illegal transition is reported as its own outcome, not as an exception,
      and leaves no event behind
- [ ] A product that does not exist is reported as its own outcome
- [ ] **Moving to `active` is refused.** `active` means "in an owner's hands"
      and ownership does not exist until v0.5.0. Per ADR 0008 and the roadmap it
      must become a consequence of ownership being established, never a label
      somebody sets
- [ ] The read is locked for update, so two concurrent changes cannot both
      compute their `previous_state` from the same row

## Risk

`risk:high`. This decides what a passport is able to say about why a product is
untrustworthy — and a suspension recorded without a reason cannot be explained
afterwards, because the history is append-only.

clevervi:
Closes #56.

The gap [`product-lifecycle.md`](docs/product-lifecycle.md) admitted: a status
change left no record of **why**. A theft report, a fraud investigation and a
disputed warranty claim all rendered as the same word, and they are not remotely
the same news for someone about to buy.

## What a history now reads like

From a live run against a reset database — registered over HTTP, then moved
through its life:

```
2026-08-14  product_suspended   theft_report       registered -> suspended   by authority
2026-09-17  product_registered  -                  -          -> -           by issuer
2026-09-17  product_reinstated  dispute_resolved   suspended  -> registered   by authority
2026-09-17  product_retired     end_of_life        registered -> retired      by issuer

after retire -> illegal_transition (retired -> suspended)
```

## Three decisions worth arguing with

**The event type reads both ends, not the destination.** `draft → registered` is
an issuer committing for the first time; `suspended → registered` is a
suspension being lifted. Recording both as "registered" would lose the fact that
something was once wrong — the fact a buyer most needs.

**`active` is absent from the input type**, so the *compiler* refuses it rather
than a reviewer. It means "in an owner's hands", ownership does not exist until
v0.5.0, and setting it today would assert somebody owns a product when nothing
has established that. ADR 0008 and the roadmap both require it to become a
consequence of ownership rather than a label anyone can set.

**The trigger stays the authority on legality.** The repository catches `TP001`
rather than re-deciding which moves are legal in TypeScript, so there is one
answer to that question instead of two that drift apart.

## Mutation checks

| Mutation | Result |
| --- | --- |
| Record a reinstatement as a registration | 2 tests red |
| Drop `previous_state` from the event | 10 tests red |
| Write an event for a no-op change | 1 test red |
| Ignore when-it-happened, always use now | 1 test red |
| **Remove the `FOR UPDATE` row lock** | **0 tests red** |

## The mutation that did not fail, and what I did about it

Removing the row lock left every test green — including the one I had just
written to cover it.

Two calls issued with `Promise.allSettled` are fast enough to serialise on their
own. Measured from the recorded timestamps: **13ms apart**, so the second
transaction's read happened after the first had committed and the race never
occurred.

I could have left it and claimed concurrency coverage. Instead the test is
renamed to what it actually proves — that each event starts where the previous
one ended, under whatever interleaving occurs — and the comment says plainly
that **the lock is defensive and has no test that can fail when it is removed**,
because reproducing the interleaving deterministically needs transaction control
this function deliberately does not expose.

A test that cannot fail is worse than no test, because it reads like coverage.
This is the second one this week, after the quiet-zone tautology in #40.

## An observation, not a defect

The run above shows a theft dated **before** the product's registration event.
That is my test data, and it is also legitimate: per ADR 0007 a record can begin
after the object exists, so an event may describe something from before
TrustPass knew about it. Ordering by `occurred_at` puts it first, which is
correct — it is where a reader would look for it.

`pnpm db:reset`, register over HTTP, then suspend, reinstate, no-op, retire, and
one illegal move — all against the built API and real Postgres.

```
packages/db  213 passed     with Postgres:    349 passed
apps/api      60 passed     without Postgres: 205 passed, 144 skipped cleanly
apps/web      76 passed
```

## Deliberately not here

- **An HTTP endpoint.** Nothing outside should move a status before
  authentication exists (`TP-141`)
- **Rendering the reason on the passport** — `TP-034`
- **The database guarantee that no status change can happen without an event** —
  #54, broadened to cover creation and status change together, with the blast
  radius measured: the two largest affected suites exist to test one thing each,
  and forcing them to record provenance would make them assert two

## Checklist

- [x] `pnpm test` — with and without a database
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Verified over HTTP against a reset database, not only a compile


packages/db/src/repositories/product-status-repository.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Which event a move produces, read from **both** ends.
+ *
+ * The destination alone is not enough: `draft → registered` is an issuer
+ * committing to a record for the first time, while `suspended → registered` is
+ * a suspension being lifted. Recording both as "registered" would lose the fact
+ * that something was once wrong, which is the fact a reader most needs.
+ */
+function eventFor(from: ProductStatus, to: ChangeableStatus): LifecycleEventType {
```
clevervi, line 70: Reading both ends is the point. Destination alone would record a lifted suspension as just another registration, losing the fact that something was once wrong — which is exactly what a buyer needs to see.

packages/db/src/repositories/product-status-repository.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Statuses this function will move a product to.
+ *
+ * `active` is deliberately absent. It means "in an owner's hands", and
+ * ownership does not exist until v0.5.0 — so setting it today would be the
+ * system asserting somebody owns a product when nothing has established that.
+ * ADR 0008 and the roadmap both require it to become a consequence of ownership
+ * rather than a label anyone can set, and leaving it out of this type means the
+ * compiler enforces that rather than a reviewer.
+ */
+export type ChangeableStatus = Exclude<ProductStatus, "active" | "draft">;
```
clevervi, line 24: Excluding active from the type means the compiler refuses it, not a reviewer. It means "in an owner's hands" and nothing establishes ownership yet, so today it would be the system asserting something nobody checked.

packages/db/src/repositories/product-status-repository.ts
```diff
@@ -0,0 +1,173 @@
...
+    return await db.transaction(async (tx) => {
+      // Locked for update, so two concurrent changes cannot both read the same
+      // `previous_state` and each record a move from it. The database trigger
+      // would still refuse an illegal result, but the losing writer would have
+      // recorded an event describing a move that never happened.
+      const [current] = await tx
+        .select({ id: product.id, status: product.status })
+        .from(product)
+        .where(eq(product.id, input.productId))
+        .for("update");
```
clevervi, line 103: This lock survives with no test that can fail when it is removed — the two calls serialise on their own at ~13ms apart. Flagged rather than dressed up as coverage.

PR #61, opened 2026-09-18
feat: show what has been recorded about a product, without concluding anything

### LINKED ISSUE #60: [PASSPORT] TP-034 Show a product's history on its passport
## Objective

Make the append-only history visible. The passport stops being a photograph of
the current state and becomes a record of what was reported, by whom, and when.

Four PRs built the spine — #52, #55, #57 and the semantics in #58/#59. Nothing a
reader can see has changed yet.

## Scope

**In scope:** the passport response carries the product's history; the page
renders it.

**Out of scope:** origin (`TP-046`), ownership events (`TP-060`+), evidence
beyond the opaque `source_reference` already stored.

## The rule that makes this `risk:critical`

**An event must never be rendered as a verdict.**

```
PRODUCT_SUSPENDED + reason theft_report
```

may be shown as *"Suspended — a theft report was recorded"*.

It may **not** become **❌ STOLEN**.

The first says who reported what. The second asserts a fact about the world
that TrustPass has not established and cannot. Per ADR 0008 an event records
that an actor reported a change; turning it into a conclusion is ADR 0003's
failure reached through the history section instead of through a badge — and it
would accuse a seller on the strength of one unverified report.

## Acceptance criteria

- [ ] The passport response includes the product's history
- [ ] **No internal key crosses the boundary** — the product's `id` must not
      leave the db package to fetch it (ADR 0005)
- [ ] Each entry shows **what happened**, **when it occurred**, **when TrustPass
      recorded it**, **the actor's capacity**, and **the reason where there is
      one**
- [ ] Both dates are shown when they differ, and the gap is presented as
      provenance, never as a reliability judgement (ADR 0008 as amended)
- [ ] Every event type and reason has wording that attributes rather than
      concludes — "a theft report was recorded", not "stolen"
- [ ] An unrecognised event type or reason renders neutrally and never as
      something worse or better than it is
- [ ] The history states an explicit **unknown period** before the first event:
      what came before is outside TrustPass, not absent from the product
- [ ] A product with no history renders the page without it, rather than failing
- [ ] Ordered by when things happened, newest first

## Cases to verify end to end

| Case | Must show |
| --- | --- |
| Issuer-registered product | the registration, its actor capacity, no reason |
| Suspended then reinstated | both events, both reasons, and **no verdict** |
| Backdated event | occurred and recorded dates both, distinctly |
| Product with no events | the passport, minus the history section |

A holder-enrolled product cannot be produced end to end until `TP-046`, so the
rendering is exercised against a `holder`-capacity event written directly.

## Risk

`risk:critical`. `CONTRIBUTING.md` defines that as a mistake corrupting the trust
model itself, and rendering a report as a conclusion is precisely that.

clevervi:
Closes #60.

Four PRs built the event spine — #52, #55, #57 and the semantics in #58/#59 —
and nothing a reader could see had changed. This makes it visible. The passport
stops being a photograph of the current state and becomes a record of what was
reported, by whom, and when.

## What a reader now sees

Rendered from a production build against a real database:

```
What has been recorded

Suspension lifted                                September 18, 2026
  Recorded by an authority because a dispute was resolved.

Registered with TrustPass                        September 18, 2026
  Recorded by the issuer.

Suspended                                        August 14, 2026
  Recorded by an authority because a theft was reported.
  Reported to have happened on August 14, 2026;
  recorded by TrustPass on September 18, 2026.

Before August 14, 2026, nothing is known to TrustPass.
The product existed; this system did not have a record of it.
```

## The line this PR exists to hold

**An event must never be rendered as a verdict.**

`PRODUCT_SUSPENDED` + `theft_report` reads *"a theft was reported"*. It does not
read **STOLEN**, and it does not get a red badge — colouring a report red would
do with styling exactly what the wording refuses to do with words.

The difference is not stylistic. The first says who filed what; the second
asserts a fact about the world TrustPass has not established and cannot, and
accuses a seller on the strength of one unverified filing. Per ADR 0008 an event
records that an actor *reported* a change.

Checked against the rendered page, not the source:

```
stolen ....... absent        fake ......... absent
counterfeit .. absent        fraudulent ... absent
```

## Mutation checks

| Mutation | Result |
| --- | --- |
| `theft_report` worded as "stolen" | 1 test red |
| `counterfeit_report` worded as a finding | 1 test red |
| Suspension coloured `negative` instead of `caution` | 1 test red |
| Unknown reason silently dropped | 1 test red |
| Unknown actor dropped | 1 test red |
| Malformed history accepted by the fetch guard | **0 red** → fixed, then 2 red |

That last row again: I added the guard and no test for it. Caught by breaking
it, covered, then re-broken to confirm the cover holds. Third time this pattern
has appeared — a guard written without a test that can fail is the recurring
mistake here, and mutation checking is the only thing that has caught it.

## Three decisions worth arguing with

**No section at all when there is no history.** A heading over nothing invites
the reading that nothing happened, when what is true is that nothing was
*recorded*. Different facts.

**The unknown period is stated, not left blank.** Per ADR 0007 as amended what
began is TrustPass's record, not the product — so the page says the product
existed and this system did not know about it, rather than letting a history
that simply stops imply the product has no earlier life.

**Both dates, with no judgement about the gap.** ADR 0008 as amended: a late
record may be the one backed by an official document and an immediate one may be
a seller's unchecked word. The page shows the distance and does not weigh it.

## ADR 0005 stayed structural

Fetching history needs the product's internal key to join on, and that key must
never leave the db package. So the join lives there —
`findHistoryByTrustPassId` takes the public identifier. A caller able to pass an
`id` in would be a caller that had already been given one.

Verified against the live response: `"id"`, `issuerId`, `productId` — **none
present**.

## Tests

```
apps/web      111 passed     with Postgres:    387 passed
packages/db   216 passed     without Postgres: 243 passed, 144 skipped
apps/api       60 passed
```

Every reason and event type the API can send is enumerated in the wording suite,
so adding one to the enum without writing words for it turns a test red.

## Deliberately not here

`origin` (`TP-046`) — so a passport cannot yet say a record began with a holder
rather than a manufacturer. Every product today is issuer-registered; the
rendering is exercised against a `holder`-capacity event and will carry origin
when it exists.

## Checklist

- [x] Verified against a production build and a reset database, over HTTP
- [x] Rendered output checked for verdict language, not just the source


apps/web/app/trustpass/claim-wording.ts
```diff
@@ -108,3 +108,86 @@ const STATUSES: Readonly<Record<string, { readonly label: string; readonly tone:
...
+};
+
+/**
+ * Reasons, worded as reports rather than findings.
+ *
+ * Read each one and ask whether it could be mistaken for a conclusion. "Theft
+ * reported" cannot; "Stolen" could.
+ */
+const EVENT_REASONS: Readonly<Record<string, string>> = {
+  theft_report: "a theft was reported",
```
clevervi, line 142: Read each reason and ask whether it could be mistaken for a conclusion. "a theft was reported" cannot; "stolen" could. That is the whole test for anything added to this map later.

apps/web/app/trustpass/claim-wording.ts
```diff
@@ -108,3 +108,86 @@ const STATUSES: Readonly<Record<string, { readonly label: string; readonly tone:
...
+  actorKind: string,
+): HistoryWording {
+  return {
+    title: EVENT_TITLES[type] ?? humanise(type),
+    because: reason ? (EVENT_REASONS[reason] ?? humanise(reason).toLowerCase()) : "",
+    actor: ACTOR_CAPACITIES[actorKind] ?? humanise(actorKind),
+    // Caution, never negative: a suspension is a report worth reading carefully,
+    // and colouring it as a finding would do with styling what the wording
+    // refuses to do with words.
+    tone: type === "product_suspended" ? "caution" : "neutral",
```
clevervi, line 191: caution, not negative, and deliberately. A red badge would turn a report into a verdict through styling after the wording refused to do it through words.

packages/db/src/repositories/lifecycle-event-repository.ts
```diff
@@ -56,3 +57,36 @@ export async function findProductHistory(
...
+ * and that is not an optimisation. The product's internal key is what links the
+ * two, and per ADR 0005 it must never leave this package — so the join has to
+ * happen here. A caller that could pass an `id` in would be a caller that had
+ * already been given one.
+ *
+ * An empty array means either no history or no such product. The passport read
+ * has already established which, so distinguishing them here would only invite
+ * a second lookup to answer a question nobody asked.
+ */
+export async function findHistoryByTrustPassId(
```
clevervi, line 74: The join is here so the internal key never leaves the package. A signature taking a product id would mean the caller already had one, which is the leak ADR 0005 forbids.

Issue #64, opened 2026-09-18
[IDENTITY] TP-047 Enrol a product that has no issuer


clevervi:
## Objective

Let a product that no business ever registered enter TrustPass.

**This is the item that unblocks testing against real hardware.** Everything
built so far can only be exercised through an issuer holding a national
registration number.

## The blocker, measured

`product.issuer_id` is **`NOT NULL`**:

```
column_name | is_nullable
issuer_id   | NO
```

So a product must belong to a registered business. A person enrolling a device
they hold is not one, and [ADR 0007](../blob/develop/docs/adr/0007-identity-may-begin-after-manufacture.md)
explicitly refuses to let them register as an issuer with an invented
registration number — that destroys the field ADR 0006 chose *because a national
authority guarantees it*.

## Scope

**In scope:** `issuer_id` becomes nullable; `record_enrolled` is written with a
`holder` actor; the passport renders a product with no issuer honestly.

**Out of scope:** who the holder is. Identifying a person needs authentication
(`TP-141`). Until then the system records the **capacity** and stops there.

## Acceptance criteria

- [ ] A product can exist with `origin: holder` and **no issuer**
- [ ] Enrolment writes `record_enrolled` with actor capacity `holder`
- [ ] **A holder enrolment never produces `active`** (ADR 0008). Asserted by test
- [ ] The live-serial unique index still means something without an issuer —
      it is currently scoped to one, so decide and test what uniqueness applies
- [ ] The passport renders without an issuer panel rather than with an empty one
- [ ] The `issuer` claim reports honestly that there is no issuer, never that one
      is unverified — those are different facts
- [ ] The unknown period before enrolment is stated
- [ ] Every guard mutation-checked, each with a test that dies when it is removed

## What must not happen

A holder-enrolled passport must claim **almost nothing**. It records that
somebody entered a serial. It is not provenance, not proof of ownership, and not
evidence about the object — and it must not be mistakable for a
manufacturer-registered record at the moment a buyer is deciding.

## Depends on

`TP-046` (origin). Reconciliation with a later manufacturer registration is #49
and must be decided before first-party registration ships.

## Risk

`risk:critical`. This is the path most products will take, and a record that
overstates what it knows about them corrupts the trust model at its widest point.

PR #65, opened 2026-09-18
feat: record where a product's TrustPass record began

### LINKED ISSUE #63: [IDENTITY] TP-046 Record where a product's TrustPass record came from
## Objective

Give every product record an **origin** — where its TrustPass record started —
and show it on the passport.

Per [ADR 0007](../blob/develop/docs/adr/0007-identity-may-begin-after-manufacture.md)
as amended: what begins is the record, not the product. Origin says where that
record began, and per [ADR 0008](../blob/develop/docs/adr/0008-events-record-what-happened-claims-assert-what-is-true.md)
it is a property of the record, set once — not the actor capacity of any one
event.

## Why it matters even before holder enrolment

`registered` now means "a TrustPass record exists" and deliberately does not say
who vouches for it (#58). **Origin is where that missing half of the answer
lives.** Without it the status is honest but uninformative, and a reader cannot
tell a factory registration from a record somebody started later.

## Scope

**In scope:** the `origin` column, set on registration, carried across the API
boundary, rendered on the passport.

**Out of scope:** holder enrolment itself — `issuer_id` is `NOT NULL` today, so a
product cannot exist without an issuer. That is `TP-047`.

## Acceptance criteria

- [ ] `product.origin` is a Postgres enum: `manufacturer`, `supply_chain`,
      `holder`
- [ ] It crosses the API boundary as part of the passport contract, not as a
      rendering decision — a marketplace reading `GET /passports/{id}` must be
      able to tell them apart (ADR 0008)
- [ ] The passport states the origin in words that **attribute rather than
      certify**: a manufacturer origin is where the record began, not proof the
      product is genuine
- [ ] An unrecognised origin renders neutrally, never as better than known
- [ ] Existing rows get a defensible default, and the migration says which and why
- [ ] **Every guard ships with a test that dies when it is removed** (CONTRIBUTING)

## The distinction this must not blur

"I can register a product" is not "I can prove this product is authentic."
`origin: manufacturer` says a manufacturer started the record. It does not say
the object in someone's hands is the one that record describes — ADR 0003, and
the reason the serial claim stays `recorded` rather than `verified`.

## Risk

`risk:high`. Origin is what a reader will use to weigh everything else on the
page, and an origin that reads as a guarantee undoes the honesty the rest of the
passport is built on.

clevervi:
Closes #63.

#58 made `registered` mean "a TrustPass record exists" and deliberately stop
short of saying who vouches for it. **Origin is where that missing half of the
answer lives** — without it the status is honest but uninformative, and a reader
cannot tell a factory registration from a record somebody started later.

## What a reader now sees

```
Record started    Started in the supply chain
                  A distributor or retailer opened this record somewhere
                  between the factory and a buyer. What happened before
                  that is outside TrustPass.
```

And in the API, because a marketplace has to be able to tell these apart without
scraping a page:

```json
"origin": "supply_chain"
```

## The word that needed the most care

`manufacturer` is the most reassuring value on the whole passport, and a reader
weighs everything else against it. So it says what it actually means:

> The maker opened this record. That places its start at the factory; **it does
> not confirm that the object you are holding is the unit it describes.**

That is ADR 0003, and the reason the serial claim stays `recorded` rather than
`verified`. "I can register a product" is not "I can prove this product is
authentic", and origin is exactly where that distinction is easiest to lose.

**No origin is toned positive.** Colouring the best one green would turn it into
the badge this project refuses to render.

## The default is a claim, so it is argued

Existing rows default to `supply_chain`, not `manufacturer`. Every product
registered before this column existed came through `POST /products` from a
business whose relationship to the factory nobody recorded — calling those
`manufacturer` would assert something no row supports.

## Mutation checks

| Mutation | Result |
| --- | --- |
| `manufacturer` origin worded as proof about the object | 1 test red |
| `holder` origin implying ownership and provenance | 1 test red |
| Unrecognised origin presented as reassuring | 1 test red |
| Database default changed to `manufacturer` | 1 test red |
| `origin` enum widened to free text | 1 test red |
| `origin` dropped from the repository projection | typecheck fails |

**One mutation was invalid and I am recording it rather than counting it.**
Changing `.default()` in the TypeScript schema left every test green — correctly,
because the real `DEFAULT` lives in the migration and the TS declaration does not
change database behaviour on its own. The mutation was re-run against the
database, where it belongs, and turned a test red.

## A repair the work uncovered

`drizzle-kit generate` refused to run:

```
[0004_snapshot.json, 0005_snapshot.json] are pointing to a parent snapshot ...
which is a collision
```

The hand-written `0005` migration from #52 had its snapshot copied from `0004`
and kept the same `id`, so both claimed the same parent. Fixed by giving `0005`
its own identity and chaining it after `0004`. It would have blocked the next
migration whoever wrote it.

Production build, reset database, registered over HTTP:

```
origin in the API response ... "supply_chain"
internal keys ................ none
rendered page ................ the row and its qualification, both present
```

A page-wide grep for "genuine" and "authentic" hits twice — both in existing
**denials**: *"It is not a certificate of authenticity"* and *"not whether the
object is genuine"*. The origin strings are asserted specifically in the unit
tests, which is where that check belongs.

## What this does not do

**It does not unblock enrolling a product that has no issuer.** `issuer_id` is
`NOT NULL`, measured rather than assumed:

```
column_name | is_nullable
issuer_id   | NO
```

So a person cannot enrol a device they hold — they would have to be a registered
business, which ADR 0007 refuses. **That is #64 (`TP-047`), and it is the item
that unblocks testing against real hardware.** This PR is its foundation.

## Tests

```
apps/web     117 passed      with Postgres:    398 passed
packages/db  221 passed      without Postgres: 253 passed, 145 skipped
apps/api      60 passed
```

## Checklist

- [x] Verified over HTTP against a production build, not only a compile


apps/web/app/trustpass/claim-wording.ts
```diff
@@ -191,3 +191,49 @@ export function describeHistoryEntry(
...
+ *
+ * None of these is toned positive. An origin is a fact about the record's
+ * provenance, not a verification of anything, and colouring the best one green
+ * would turn it into the badge this project refuses to render.
+ */
+const ORIGINS: Readonly<Record<string, { readonly label: string; readonly detail: string }>> = {
+  manufacturer: {
+    label: "Started by the manufacturer",
+    detail:
+      "The maker opened this record. That places its start at the factory; it does not confirm that the object you are holding is the unit it describes.",
```
clevervi, line 213: The qualification is the point of this string, not a caveat on it. Without "does not confirm that the object you are holding is the unit it describes", a manufacturer origin reads as proof about the object — which is the badge ADR 0003 refuses.

packages/db/src/schema/product.ts
```diff
@@ -111,6 +135,17 @@ export const product = pgTable(
...
+    /**
+     * Where this record started. See `productOrigin`.
+     *
+     * Defaulted to `supply_chain` rather than `manufacturer`, and the default is
+     * a claim like any other: every product registered before this column
+     * existed came through `POST /products` from a business whose relationship
+     * to the factory nobody recorded. Calling those `manufacturer` would assert
+     * something no row supports.
+     */
+    origin: productOrigin("origin").notNull().default("supply_chain"),
```
clevervi, line 147: supply_chain, not manufacturer, and the choice is a claim. Rows written before this column existed came from businesses whose relationship to the factory nobody recorded; defaulting them to manufacturer would assert something no row supports.

Issue #49, opened 2026-09-17
[IDENTITY] Reconcile a holder enrolment with a later manufacturer registration


clevervi:
## Why this exists as an issue and not a paragraph

[ADR 0007](../blob/develop/docs/adr/0007-identity-may-begin-after-manufacture.md)
records that this problem becomes reachable, and deliberately does not solve it.
An acknowledged problem with no owner is a problem that disappears.

**This must be decided before first-party registration ships, not after.**

## The situation

A person enrols a graphics card they hold. Later, its manufacturer registers the
same serial.

The live-identity index is per issuer —
`product_live_issuer_serial_idx` on `(issuer_id, lower(serial))` where
`status <> 'retired'` — so **both records exist and neither constraint fires**.

The system now holds two TrustPass IDs for one physical object.

## Why neither obvious answer works

**Merge them.** Two records become one, and the holder's TrustPass ID either
survives or dies. If it dies, any QR already printed or shared resolves to
nothing — a code on a physical object that stops working is the failure the
whole identity design exists to prevent. If it survives, the manufacturer's
record is subordinated to one created by a stranger.

**Leave them separate.** Two passports for one object, each telling a partial
story, and a buyer scanning one has no idea the other exists. That is worse than
no record: it looks complete.

**Let the manufacturer claim it.** Closest to right, and it needs an answer to
"how does a manufacturer prove this is the unit they made?" — which is the same
serial-matching problem, from the other side.

## What makes this hard rather than tedious

The two records are not duplicates. They are **two different assertions about
the same object**, made by parties with different standing, covering different
periods. Any resolution has to preserve:

- both TrustPass IDs resolving to something truthful, forever
- who asserted what, and when — neither record may absorb the other's claims
- the unknown period on the holder record, which does not vanish because a
  manufacturer later appeared

## Out of scope

Deciding it here. This issue exists so the decision is scheduled and visible.
The outcome belongs in an ADR.

## Depends on

- Lifecycle events (`TP-050`…`TP-053`) — any resolution is itself an event
- Authentication (`TP-141`) — "the manufacturer claims it" means nothing without
  knowing who is asking

PR #66, opened 2026-09-18
feat: enrol a product that no business ever registered

### LINKED ISSUE #64: [IDENTITY] TP-047 Enrol a product that has no issuer

clevervi:
Closes #64.

**This is the item that unblocks testing against real hardware.** Everything
built so far could only be exercised through an issuer holding a national
registration number.

## The blocker, and why the obvious fix was refused

```
column_name | is_nullable
issuer_id   | NO
```

Per ADR 0007 a person enrolling a device they hold is not an issuer and must not
become one with an invented registration number — that destroys the field ADR
0006 chose *because a national authority guarantees it*.

So `issuer_id` is nullable, and a check ties its absence to a holder origin:
**never both, never neither.**

## The part that needed deciding rather than inheriting

Review raised this and it was the right thing to raise. **Postgres treats NULL
as distinct from NULL**, so the moment `issuer_id` can be null, every holder
record falls outside the existing unique index.

Measured, not assumed:

```
three rows, NULL issuer, same serial -> all three inserted cleanly
```

The index stopped enforcing anything for exactly the records this PR creates.

### The policy, and the option not taken

`NULLS NOT DISTINCT` would fix it, and Drizzle 0.45.2 cannot express it — but
that is the smaller reason. **The policy for a record with no issuer is a
different policy, and it should read as one** instead of hiding in how Postgres
compares nulls. So it is a separate partial index:

```sql
CREATE UNIQUE INDEX product_live_holder_serial_idx ON product (lower(serial))
WHERE issuer_id IS NULL AND status <> 'retired';
```

Verified against Postgres, all four cases:

| Case | Result |
| --- | --- |
| Holder record, no issuer | accepted |
| **Second live holder record, same serial** | **refused** |
| Issuer record with the same serial | accepted — serials are only unique inside a maker's numbering |
| Retire, then re-enrol | accepted — retirement releases the serial |

**Why one live holder record per serial:** otherwise anyone could enrol a
device, have something unwelcome recorded against it, enrol it again and show
the clean passport. That breaks the single rule the serial indexes exist for.

**What it costs:** two genuinely different products with colliding serials from
different makers, both holder-enrolled — the second is refused. A real false
positive, the cheaper of the two failures, and #49 owns resolving a collision
properly.

## Invariants, each verified against real Postgres

| Invariant | Result |
| --- | --- |
| Holder record with no issuer | accepted |
| **Holder record born `active`** | **refused** — `TP001`, by the existing status trigger |
| Holder origin naming an issuer | refused — check violation |
| Issuer origin with no issuer | refused — check violation |

The `active` one came free: the trigger from `TP-025` already forbids birth as
`active`, and `changeProductStatus` already excludes it from its type. Entering
a serial establishes nothing about who holds the object.

## A left join, not an inner one

An inner join drops a holder record entirely and reports an existing product as
**not found** — the "we could not check" failure reached by way of a join.

## "No issuer" is not "unverified issuer"

The distinction review asked for, carried the whole way:

- The claim is `not_present`, not `unverified`
- Null crosses the wire, and the OpenAPI schema says why
- The page renders **no issuer panel at all** rather than one with blanks — an
  empty Issuer card reads as an issuer whose details are missing
- The banner does not reuse the unverified wording, which asserts a company exists

## Mutation checks

| Mutation | Result |
| --- | --- |
| Drop the holder serial index | 1 test red |
| Drop the origin/issuer check | 3 tests red |
| Report no issuer as `unverified` | 1 test red |
| Remove the "no issuer" wording | 1 test red |

## Two pre-existing tests failed, and both failures were correct

`requires an issuer` asserted `NOT NULL`. That is no longer true, so it now
asserts the check constraint. **The rule did not weaken, it moved** — and it now
also catches a holder record that names an issuer, which `NOT NULL` never did.

The origin test from #65 inserted `origin: "holder"` with an issuer, which the
new check refuses. Split into the two cases it was conflating.

## Verified end to end

A product enrolled with no issuer, read back over HTTP from a production build:

```
issuer       : null           origin : "holder"
status       : "registered"   claim  : issuer -> not_present

no issuer panel ........................ ok
says there is no company ............... ok
does not claim an unverified issuer .... ok
origin shown ........................... ok
status Registered, not Active .......... ok
```

```
apps/web     119 passed      with Postgres:    409 passed
packages/db  230 passed
apps/api      60 passed
```

## Checklist

- [x] Verified over HTTP against a production build, not only a compile


packages/db/src/schema/product.ts
```diff
@@ -191,10 +192,44 @@ export const product = pgTable(
...
+    // Without it anyone could enrol a device, have something unwelcome recorded
+    // against it, enrol it again and show the clean passport -- which breaks the
+    // one rule the serial indexes exist for. A holder has no issuer namespace to
+    // scope uniqueness to, so the scope is every holder record.
+    //
+    // It costs a real false positive: two genuinely different products with
+    // colliding serials from different makers, both holder-enrolled, and the
+    // second is refused. That is the cheaper failure of the two, and #49 owns
+    // resolving a collision properly.
+    uniqueIndex("product_live_holder_serial_idx")
```
clevervi, line 220: A separate index rather than NULLS NOT DISTINCT, and not mainly because Drizzle cannot express it: the policy for a record with no issuer is a different policy and should read as one instead of hiding in how Postgres compares nulls.

packages/db/src/schema/product.ts
```diff
@@ -191,10 +192,44 @@ export const product = pgTable(
...
+      .on(sql`lower(${table.serial})`)
+      .where(sql`${table.issuerId} IS NULL AND ${table.status} <> 'retired'`),
+
+    // A record has an issuer or it has a holder origin. Never both, never
+    // neither: an issuer-registered product with no issuer is an orphan, and a
+    // holder-enrolled one with an issuer is a person claiming a company's
+    // standing.
+    check(
+      "product_holder_has_no_issuer",
```
clevervi, line 229: Never both, never neither. NOT NULL never caught a holder record naming an issuer; this does, so the rule got stronger rather than weaker when the column opened up.

PR #68, opened 2026-09-18
feat: enrol a product you hold, over HTTP

### LINKED ISSUE #67: [IDENTITY] TP-048 Enrol a product you hold, over HTTP
## Objective

Give holder enrolment a path a person can actually use.

## Why this exists separately from TP-047

`TP-047` (#66) made the **model** support a product with no issuer — the
database, the passport, the claims. It did not give anyone a way to create one.

`POST /products` requires an `issuer` object and answers `issuer_not_found`
without it. So today a holder-enrolled product can only be made by writing to
the database directly, which is not a product.

## Scope

**In scope:** `POST /enrolments`, and the domain function behind it.

**Out of scope:** who the holder is. Identifying a person needs `TP-141`; until
then the system records the capacity and stops there.

## Why a separate path rather than making `issuer` optional on `POST /products`

Registration and enrolment are different acts making different claims. An
optional field would let a caller slide between them by omission, and the
response to "you forgot the issuer" would become "we silently made this a
holder enrolment" — a claim nobody chose.

It also matches the prefix rule from `TP-030`: when `TP-141` locks writes,
`/products` and `/enrolments` may need different authority, and a prefix is
easier to reason about than a per-field exception.

## Acceptance criteria

- [ ] `POST /enrolments` creates a product with `origin: holder`, no issuer, and
      status `registered`
- [ ] The product and its `record_enrolled` event are written in **one
      transaction**, with actor capacity `holder`
- [ ] **It never produces `active`.** Entering a serial establishes nothing
      about who holds the object
- [ ] A second live enrolment of the same serial answers **409**, naming the
      TrustPass ID that already holds it — so a caller that timed out and
      retried learns the identifier rather than creating a second passport
- [ ] The response carries no internal key (ADR 0005)
- [ ] The OpenAPI document states plainly that enrolling claims nothing about
      the object, and does not describe the result as verification
- [ ] Every guard mutation-checked, each with a test that dies when it is removed

## The sentence this endpoint must not imply

"I enrolled this product" is not "this product is mine" and not "this product is
genuine". It is: **somebody entered this serial.** The response, the OpenAPI
description and the passport must all be readable only that way.

## Risk

`risk:high`. This is the path most products will take into the system, and the
one most likely to be mistaken for a claim of ownership.

clevervi:
Closes #67.

## A correction first

I said #66 meant a real product could now be enrolled. **That was wrong.**

`TP-047` made the *model* support a record with no issuer. It gave nobody a way
to create one: `POST /products` requires an `issuer` object and answers
`issuer_not_found` without it, so a holder-enrolled product could only be made
by writing to the database by hand. That is not a product.

This is the part that was missing.

## It works

```
POST /enrolments  { "brand": "ASUS", "model": "ROG Strix RTX 5070 Ti OC",
                    "serial": "M1LMCS004896", "category": "gpu" }

201  { "trustpassId": "TP1-1NR49F78XESWR768STP58S41PJY",
       "status": "registered", "origin": "holder", ... }

the same serial again
409  { "error": "duplicate_serial",
       "message": "Serial M1LMCS004896 is already enrolled as TP1-1NR49F78..." }
```

And its passport, rendered from a production build:

```
ASUS ROG Strix RTX 5070 Ti OC

  No business registered this product
  Somebody who had the product opened this record. Nothing here comes from
  a manufacturer, a distributor or a shop, and nobody has vouched for it.

  Status          Registered
  Serial number   ••••4896 — 8 characters withheld
  Record started  Started by whoever held it

  Issuer          None — no business registered this product ... which is
                  not the same as a company nobody has checked yet
  Serial number   Recorded, not checked
  Physical product  Not independently verified
```

## Why a separate path, not an optional `issuer`

Registration and enrolment are different acts making different claims. An
optional field lets a caller slide between them **by omission** — "you forgot
the issuer" would silently become "we made this a holder enrolment", a claim
nobody chose. And when `TP-141` locks writes, a prefix is easier to reason about
than a per-field exception.

`status` and `origin` are not parameters either, for the same reason: they are
not choices.

## A defect found by reading the page, not by a test

The serial claim said **"The issuer supplied this serial."** Since `TP-047` a
record may have no issuer at all, so that sentence **invented one on every
holder enrolment** — the same mistake as reporting a missing issuer as an
unverified one, which this project had just spent a PR removing.

Caught by rendering the passport of an enrolled product and reading it. There is
a test now.

## Mutation checks

| Mutation | Result |
| --- | --- |
| Enrolment made `active` | 7 tests red |
| Origin changed to `supply_chain` | 7 tests red |
| Actor recorded as the system, not a holder | 1 test red |
| Any unique violation read as a duplicate serial | **0 red** → covered, then 1 red |
| Serial lookup made case sensitive | 1 test red |
| Serial attributed to an issuer again | 2 tests red |
| "Nothing has compared it to the object" removed | 1 test red |

**The fourth row is the recurring mistake, caught by the rule added in #62.**
I wrote the repository and no integration test for it, so three guards sat
uncovered. The rule says a guard ships with a test that dies without it; this is
the first time the rule caught me rather than a reviewer.

The specific guard: the duplicate path checks the **constraint name**. Reporting
a TrustPass ID collision as "already enrolled" would tell a caller their serial
is taken when it is not — sending them to change the one thing that was right.

## What this endpoint must not be read as

"I enrolled this product" is not "this product is mine" and not "this product is
genuine". It is: **somebody entered this serial.** The response, the OpenAPI
description and the passport are all written to be readable only that way.

## Tests

```
apps/web     121 passed      with Postgres: 420 passed
packages/db  239 passed
apps/api      60 passed
```

## Checklist

- [x] Verified over HTTP against a production build and a reset database


apps/api/src/routes/enrolments.ts
```diff
@@ -0,0 +1,110 @@
+import { createRoute, type OpenAPIHono, z } from "@hono/zod-openapi";
+import type { AppDependencies } from "../dependencies.js";
+import { ApiErrorCode, ApiErrorSchema } from "../http/errors.js";
+import { ProductCategorySchema } from "./product-category.js";
+
+/**
+ * Enrolling a product you hold.
+ *
+ * A separate path from `POST /products` rather than an optional issuer on it.
```
clevervi, line 9: status and origin are not parameters because they are not choices. An optional issuer on POST /products would let a caller slide into a holder enrolment by forgetting a field — a claim nobody made.

packages/db/src/repositories/enrolment-repository.ts
```diff
@@ -0,0 +1,100 @@
...
+        // slightly before the row it describes.
+        occurredAt: sql`now()`,
+      });
+
+      return { ok: true, product: created };
+    });
+  } catch (error) {
+    const violation = violatedConstraint(error);
+
+    if (violation.code === UNIQUE_VIOLATION && violation.constraint === LIVE_HOLDER_SERIAL_INDEX) {
```
clevervi, line 68: Checking the constraint name, not just the SQLSTATE. A TrustPass ID collision reported as "already enrolled" would tell a caller their serial is taken when it is not, sending them to change the one thing that was right.

PR #70, opened 2026-09-18
feat: limit what each capacity may record

### LINKED ISSUE #69: [LIFECYCLE] TP-053 Limit what each capacity may record
## Objective

Stop any actor from recording any event. A capacity should only be able to
assert what its standing supports.

## Why this is urgent now

`TP-048` shipped `POST /enrolments`, so **anyone can now put a record into the
system**. Nothing constrains what may then be recorded against it.

Concretely: a holder can record `product_reinstated` and clear a theft report
somebody else filed against their own device. The check constraints say the
event is well formed. Nothing says the actor had any business making it.

## Scope

**In scope:** which `actor_kind` may record which `lifecycle_event_type`,
enforced where it cannot be bypassed.

**Out of scope:** *who* the actor is — that is authentication (`TP-141`). This
constrains what a claimed capacity may assert, which is a different question and
answerable today.

## The policy to decide and write down

| Capacity | May record | Reasoning |
| --- | --- | --- |
| `holder` | `record_enrolled`, `product_retired` | Can say "I put this in" and "this is done". Cannot clear an accusation |
| `issuer` | `product_registered`, `product_retired` | Its own registration and end of life |
| `authority` | `product_suspended`, `product_reinstated` | A report and its resolution belong to whoever can investigate |
| `system` | `record_corrected` | Only TrustPass corrects its own record |

**The one that matters most: a holder must not be able to reinstate.** A product
suspended over a theft report, cleared by the person holding it, is the system
helping launder a stolen device.

## Acceptance criteria

- [ ] The policy is one declaration, not a condition scattered across write paths
- [ ] It is enforced **in the database**, so a path that never went through the
      application cannot bypass it — the reasoning that put status transitions
      in a trigger
- [ ] A project SQLSTATE distinguishes it from `TP001` and `TP002`
- [ ] The TypeScript declaration and the database rule are compared by a test
      that fails when they diverge — the pattern `product-status` already uses
- [ ] `changeProductStatus` and `enrolProduct` refuse a disallowed pairing as an
      outcome, not an exception

## Risk

`risk:critical`. `CONTRIBUTING.md` defines that as a mistake corrupting the trust
model itself. An actor that can record anything makes every claim on a passport
worth what the least trustworthy writer decided to type.

clevervi:
Closes #69. Authored by [@raishark](https://github.com/raishark) — see below.

## Why this could not wait

`TP-048` shipped `POST /enrolments`, so **anyone can now put a record into the
system**. Nothing constrained what could then be recorded against it.

Concretely, before this PR:

```sql
-- an authority suspends a product over a theft report
INSERT ... ('product_suspended', 'authority', 'theft_report')   -- accepted

-- the HOLDER clears it
INSERT ... ('product_reinstated', 'holder', 'dispute_resolved') -- also accepted
```

**A product suspended over a theft report, cleared by the person holding it, is
the system helping launder a stolen device.** The check constraints said the
event was well formed. Nothing said the actor had any business making it.

Now:

```
ERROR:  A holder cannot record "product_reinstated".   (TP003)
ERROR:  A issuer cannot record "product_suspended".    (TP003)
```

## The policy

| Capacity | May record | Why the gaps |
| --- | --- | --- |
| `holder` | `record_enrolled`, `product_retired` | Can say "I put this in" and "this is done". **Cannot clear an accusation** |
| `issuer` | `record_enrolled`, `product_registered`, `product_retired` | **Cannot suspend** — burying a complaint about its own product is the same conflict from the other side |
| `authority` | `product_suspended`, `product_reinstated`, `product_retired` | A report and its resolution belong together; splitting them leaves suspensions nobody can lift |
| `system` | `record_corrected` | Only TrustPass corrects TrustPass's record |

Authentication decides *who* somebody is — still `TP-141`. This decides what a
**claimed capacity** may assert, which is a different question and answerable
today.

## Declared once, enforced where it cannot be bypassed

TypeScript table plus a database trigger, and **every pairing compared against
real Postgres** so a divergence is a failing test rather than a rule that
quietly stopped applying. `TP003` is distinct from `TP001` and `TP002`.

## Mutation checks

| Mutation | Result |
| --- | --- |
| Drop the authority trigger | 15 tests red |
| Grant a holder the power to reinstate | 4 tests red |

## Seven existing tests failed, and every failure was correct

They used pairings that are now refused — an issuer suspending, a holder
recording a registration. Those tests were written when any actor could record
anything. They now use pairings the standing supports.

One was worth rewriting rather than patching: a sweep over actors with a single
fixed event type was testing the authority trigger by accident. Each capacity
now gets a type it is allowed to record.

---

## The attribution correction in this PR

The first commit restores a README line **#46 removed on a false premise.**

I wrote there that `@raishark` "has never committed", having checked with
`git log --format='%an'` — which shows the author *name*. **GitHub attributes
commits by email**, and `raivel.studio@gmail.com` belongs to Raishark.

The API is unambiguous:

```
clevervi   40 commits
Raishark    7 commits
```

And Raishark's are the **foundation**: the pnpm monorepo, the Drizzle client.
Authorship stopped at PR #1 when the repository's git identity changed — not by
any decision, and I removed the line documenting it on a check I ran wrong.

The line now also states the part that matters, which #46 got right for the
wrong reason:

> Both write code. **Neither ever reviews the other.** They are the same person,
> so an approval between them would be a review signal with nothing behind it.

`.github/CODEOWNERS` records one maintainer for exactly that reason, and the
merge bar is built around having no second reader rather than pretending to
have one.

## Tests

```
apps/web     121 passed      with Postgres: 442 passed
packages/db  261 passed
apps/api      60 passed
```

## Checklist

- [x] The refusal verified directly through `psql`, not only through the suite


packages/db/src/domain/recording-authority.ts
```diff
@@ -0,0 +1,78 @@
...
+  Record<LifecycleActorKind, readonly LifecycleEventType[]>
+> = {
+  /**
+   * Somebody who says they have the object.
+   *
+   * They can say "I put this into TrustPass" and "this has reached the end of
+   * its life". They cannot register on a business's behalf, and they cannot
+   * clear an accusation made against the thing they hold.
+   */
+  holder: ["record_enrolled", "product_retired"],
```
clevervi, line 32: The entry that matters is the one that is absent. A holder cannot reinstate — a suspension over a theft report, cleared by whoever holds the object, is the system helping launder a stolen device.

packages/db/src/domain/recording-authority.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * A registered business.
+   *
+   * Its own registration, and retiring what it registered. Suspension is not
+   * here: an issuer suspending a product to bury a complaint about it is the
+   * same conflict of interest as a holder clearing one.
+   */
+  issuer: ["record_enrolled", "product_registered", "product_retired"],
```
clevervi, line 41: Suspension is deliberately not here. An issuer suspending a product to bury a complaint about it is the same conflict of interest as a holder clearing one, seen from the other side.

PR #73, opened 2026-09-18
fix: a holder could launder a suspension by retiring the product

### LINKED ISSUE #72: [LIFECYCLE] A holder can launder a suspension by retiring the product
## The path, reproduced against Postgres

```
1. an authority suspends a product over a theft report   -> suspended
2. the HOLDER retires it themselves                      -> retired
3. retirement releases the serial, so it is re-enrolled  -> registered

   old passport   TP1-LAUNDER0...   retired      2 events (the theft is here)
   NEW passport   TP1-LAUNDER1...   registered   0 events
```

**The new passport knows nothing about the theft report.**

## Why TP-053 did not close this

`TP-053` (#70) stopped a holder recording `product_reinstated` — clearing a
suspension directly. It allows `holder → product_retired`, which is normally
correct: a person can say their own device reached end of life.

What makes it laundering is the *other* rule it meets. The live-serial index
excludes retired products **on purpose**, so a warranty replacement can reuse
the serial of the unit it replaces
([`product.ts`](../blob/develop/packages/db/src/schema/product.ts)). Retirement
frees the serial.

Neither rule is wrong. Together they are a door: **suspend → retire → re-enrol
clean.**

The append-only history survives on the old record, and that is exactly the
problem — nobody scanning the new code will ever see the old one.

## The fix

**A product may only move from `suspended` to `retired` by an authority.**

Whoever can investigate a report is who can decide the object's life ends while
it is open. A holder or an issuer must get the suspension resolved first, and
that resolution is recorded.

This keeps every legitimate case: `registered → retired` by a holder or an
issuer is untouched, so warranty replacement still works.

## Acceptance criteria

- [ ] A `product_retired` event whose `previous_state` is `suspended` is refused
      unless the actor is an `authority`
- [ ] Enforced **in the database**, not only in a write path
- [ ] `registered → retired` by a holder and by an issuer both still work
- [ ] `suspended → retired` by an authority works
- [ ] The full laundering sequence is a test, end to end, and fails at step 2

## Residual risk, to be stated rather than hidden

An **authority** retiring a suspended product still releases the serial. That is
a decision an authority is entitled to make, and it is recorded. Worth naming so
nobody believes this closes serial reuse entirely.

## Risk

`risk:critical`. This is the trust model failing at its most load-bearing point:
a stolen device getting a clean passport, produced by the system itself.

clevervi:
Closes #72. Raised by review, and worse than the review realised.

## The path, reproduced against Postgres before the fix

```
1. an authority suspends a product over a theft report   -> suspended
2. the HOLDER retires it themselves                      -> retired
3. retirement releases the serial, so it is re-enrolled  -> registered

   old passport   retired      2 events (the theft is here)
   NEW passport   registered   0 events
```

**The new passport knows nothing about the theft report.**

The review spotted that a holder could retire a suspended product and asked
whether that should be allowed. It did not connect it to the second rule, which
is what turns an odd transition into laundering: **retirement frees the serial.**

## Why #70 did not close this

`TP-053` stopped a holder recording `product_reinstated` — clearing a suspension
directly. It allows `holder → product_retired`, which is normally correct: a
person can say their own device reached end of life.

The live-serial index excludes retired products **on purpose**, so a warranty
replacement can reuse the serial of the unit it replaces.

Neither rule is wrong. Together they are a door.

## The fix

**Only an authority may move a product from `suspended` to `retired`.**

Whoever can investigate a report is who can decide the object's life ends while
it is open. A holder or an issuer must get the suspension resolved first — and
that resolution is itself recorded.

```
holder retiring a suspended product   -> TP003, "Resolve the suspension first."
issuer retiring a suspended product   -> TP003
authority retiring a suspended product -> allowed
holder retiring a registered product  -> allowed (warranty replacement)
```

## Mutation checks

| Mutation | Result |
| --- | --- |
| Remove the suspended-retirement guard from the trigger | 2 tests red |
| Make the TypeScript rule permissive | 3 tests red |

## The test runs the whole sequence, not the guard alone

Everything after step 2 — the serial released, a fresh enrolment, a passport
with no history — follows from that one step succeeding. So that is where the
test asserts, and it asserts the product is still `suspended` afterwards.

## Residual risk, named rather than hidden

**An authority retiring a suspended product still releases the serial.** That is
a decision an authority is entitled to make and it is recorded, but it means
this does not close serial reuse entirely. #49 owns the collision policy.

## On the review's other two points

**`actor_kind` is still self-declared until `TP-141`.** Correct, and the PR that
introduced it said so: this is authorization, not authentication. The API is not
exposed and the roadmap forbids exposing it before `TP-141`.

**The README discrepancy.** Checked on `develop` rather than assumed —
`git show origin/develop:README.md` shows both accounts under Maintainer. That
snapshot is stale; the merge landed.

## Tests

```
apps/web     121 passed      with Postgres: 449 passed
packages/db  268 passed
apps/api      60 passed
```

## Checklist

- [x] The path verified through `psql` before and after, not only through the suite


packages/db/src/domain/recording-authority.ts
```diff
@@ -62,6 +63,22 @@ export class UnauthorisedRecording extends Error {
...
+ * Retiring a product frees its serial — the live-serial index excludes retired
+ * rows on purpose, so a warranty replacement can reuse the serial of the unit it
+ * replaces. That is correct, and combined with a holder being able to retire it
+ * became a door: suspend over a theft report, retire, re-enrol the serial, and
+ * the new passport has no history.
+ *
+ * So whoever can investigate a report is who can decide the object's life ends
+ * while that report is open. Everything else is untouched.
+ */
+export function mayRetireFrom(actorKind: LifecycleActorKind, from: ProductStatus): boolean {
```
clevervi, line 78: Two correct rules meeting to form a door. Retirement freeing the serial is deliberate — warranty replacement needs it — and a holder retiring their own device is reasonable. It is only together that they launder a theft report.

PR #74, opened 2026-09-18
feat: guarantee provenance in the database, not in a convention

### LINKED ISSUE #54: [LIFECYCLE] Guarantee no product state can change without provenance

clevervi:
Closes #54. Authored by [@raishark](https://github.com/raishark).

`TP-051` and `TP-052` made the registration and status paths write the product
and its event together. **That covers the paths that exist.** It does nothing
about a bulk import, a migration, a fixture or a console session — and per ADR
0008 that failure is invisible, because nothing later can distinguish a history
that is *missing* from one that is *empty*.

```
INSERT INTO product ...                    -> TP004 at COMMIT
UPDATE product SET status = 'suspended'    -> TP004 at COMMIT
```

## Three decisions in the trigger

**Deferred**, because the event references the product and so must be written
after it. A plain `AFTER` trigger fires while the event cannot exist yet, which
makes the rule impossible to satisfy rather than hard to bypass.

**Scoped to this transaction, not to history.** A product moving
`registered → suspended → registered → suspended` would otherwise satisfy the
third move with the event that recorded the first — leaving the last transition
unexplained while looking accounted for. There is a test for exactly that.

**Separate triggers rather than clauses in the status guard.** That guard raises
`TP001` immediately; these fire at `COMMIT`. So an illegal move is still
reported as illegal rather than as unexplained, and a caller is not told it
forgot to explain a move it was never going to be allowed to make.

## Two rules that disagreed, found by doing this

The status machine has always allowed `registered → active`. **No event type
could record it.** The moment provenance became a guarantee, a legal transition
became impossible to perform.

Neither rule was wrong on its own, and neither was found by reading either — the
suite that walks every transition found it. `product_activated` closes the gap,
recordable only by the `system`, because ADR 0008 makes `active` a consequence
of ownership rather than a label an actor sets.

## Mutation checks

| Mutation | Result |
| --- | --- |
| Drop the insert provenance trigger | 1 test red |
| Drop the status provenance trigger | **0 red** → covered, then 2 red |
| Remove the transaction scoping from the check | 1 test red |

**The second row again.** The first run found the status trigger had no test
that could die without it. That is the guard-clause rule from #62 catching its
own author, before shipping rather than in review — which is the first time it
has worked that way round.

## The cost, paid deliberately

**29 fixtures across 7 suites** wrote products and status changes directly. Each
now goes through a helper that writes the event too, so the suites keep testing
one thing each rather than also asserting provenance.

Tests whose subject *is* provenance deliberately do **not** use the helpers:
they write the raw statement and assert the refusal.

**Teardowns that deleted products are gone.** Products carry events, events
cannot be removed, and the foreign keys are `RESTRICT` — a teardown that
succeeded would prove a product's history can be erased.

**One test lost its premise entirely.** It inserted a bare product to check the
history reader returned an empty array for one with no provenance. That product
can no longer exist, so the test asserts the closure instead of the defence.

## Two mistakes made and found while doing it

A cleanup regex stripped `.returning()` from an issuer fixture, so `issuerId`
was silently `undefined` and a product was created with no issuer. It surfaced
as a passport test failing on a missing issuer, three layers away from the
cause.

Fixtures using `new Date()` for `occurred_at` began failing the not-in-future
check. The host clock runs ahead of the container's, so a JavaScript `Date` can
be later than the transaction timestamp `recorded_at` defaults to. Every fixture
now uses `now()`, which is what the production paths already did after `TP-051`
found the same thing from the other direction.

## Tests

```
apps/web     121 passed      with Postgres:    459 passed
packages/db  278 passed      without Postgres: 252 passed, 201 skipped
apps/api      60 passed
```

## Checklist

- [x] Migrations named, forward-only, applied from a virgin volume
- [x] Error ordering verified directly through `psql`, by SQLSTATE rather than message


packages/db/drizzle/0010_provenance_is_guaranteed.sql
```diff
@@ -0,0 +1,79 @@
...
+  -- transition would go unexplained while looking accounted for.
+  --
+  -- recorded_at defaults to now(), which is the transaction timestamp, so an
+  -- event written in this transaction carries exactly this value.
+  IF NOT EXISTS (
+    SELECT 1 FROM lifecycle_event
+    WHERE product_id = NEW.id
+      AND previous_state = OLD.status
+      AND resulting_state = NEW.status
+      AND recorded_at >= transaction_timestamp()
```
clevervi, line 57: Scoped to this transaction, not to history. Without it, a product moving registered to suspended to registered to suspended satisfies the third move with the event that recorded the first — unexplained while looking accounted for.

packages/db/src/schema/lifecycle-event.ts
```diff
@@ -48,6 +48,19 @@ export const lifecycleEventType = pgEnum("lifecycle_event_type", [
...
+   * The status machine has always allowed `registered -> active`, and until now
+   * no event type could record it — so once provenance became a database
+   * guarantee, a legal transition became impossible to perform. Two rules that
+   * disagreed, found by the transition suite rather than by reading either.
+   *
+   * Only the system may record it. Per ADR 0008 `active` is a consequence of
+   * ownership being established, never a label an actor sets, and ownership
+   * arrives in v0.5.0.
+   */
+  "product_activated",
```
clevervi, line 63: Two rules that disagreed and neither was wrong alone: the status machine allowed registered to active and no event type could record it, so the moment provenance became a guarantee a legal transition became impossible.

PR #76, opened 2026-09-18
fix: one event could explain two identical status changes

### LINKED ISSUE #75: [LIFECYCLE] One event can explain two identical status changes
## The hole, reproduced against Postgres

```sql
BEGIN;
UPDATE product SET status = 'suspended'  WHERE ...;  -- move 1
UPDATE product SET status = 'registered' WHERE ...;  -- move 2
UPDATE product SET status = 'suspended'  WHERE ...;  -- move 3
INSERT INTO lifecycle_event ... registered -> suspended ...;  -- explains move 1
INSERT INTO lifecycle_event ... suspended -> registered ...;  -- explains move 2
COMMIT;                                              -- accepted
```

```
transition events: 2      status changes: 3      final status: suspended
```

**Move 3 was satisfied by the event that explained move 1.** They share a
`(previous_state, resulting_state)` pair, and the deferred check only asks
whether *some* matching event exists in this transaction.

## Why #54 did not catch it

The check scoped the search to the transaction, which stops an event from an
*earlier* transaction being reused. It does not stop an event from the *same*
transaction explaining two moves.

So the guarantee is **"some event matches"**, not **"this change has its
event"** — raised in review, and correct.

## What already works, verified rather than assumed

Two different products moved in one transaction with an event for only one of
them **is** refused. The trigger is row-level, so each product's firing looks for
its own event.

## The fix

**One status change per product per transaction.**

At commit, require **exactly one** transition event for this product in this
transaction, matching this move. Two moves then produce either two events — a
count of 2, refused — or one event, also refused.

That makes the correspondence 1:1 by construction rather than by counting, and
it needs no new column on either table.

The rule is defensible on its own: a transaction that moves a product twice is
doing two things, and each deserves its own transaction and its own recorded
reason. Every existing write path already does exactly one.

## Acceptance criteria

- [ ] Three moves and two events in one transaction is refused
- [ ] Three moves and three events in one transaction is refused — the rule is
      one move, not balanced bookkeeping
- [ ] One move and one event still works
- [ ] Two products moved with one event between them is still refused
- [ ] An illegal move still reports `TP001`, not a provenance complaint

## Risk

`risk:high`. Not reachable through any write path that exists — every one makes
a single move per transaction. But #54 exists precisely for the paths nobody has
written yet, so dismissing it would contradict the reason it was built.

clevervi:
Closes #75. Raised in review of #74, verified before being accepted.

## The hole, reproduced against Postgres

```sql
BEGIN;
UPDATE product SET status = 'suspended'  ...;  -- move 1
UPDATE product SET status = 'registered' ...;  -- move 2
UPDATE product SET status = 'suspended'  ...;  -- move 3
INSERT event registered -> suspended;          -- explains move 1
INSERT event suspended -> registered;          -- explains move 2
COMMIT;                                        -- accepted
```

```
transition events: 2     status changes: 3     final status: suspended
```

**Move 3 was satisfied by the event that explained move 1.** They share a
`(previous_state, resulting_state)` pair, and the check only asked whether
*some* matching event existed in the transaction.

So the guarantee was **"some event matches"**, not **"this change has its
event"**. The review was right and the distinction matters.

## The fix, and why it is shaped this way

Counting a trigger's own firings from inside a row trigger is not possible. So
the rule is turned around: **exactly one transition event for this product in
this transaction.**

Two moves then produce either two events — a count of 2, refused — or one event,
also refused. The correspondence becomes **1:1 by construction** rather than by
bookkeeping.

The rule stands on its own merits. A transaction that moves a product twice is
doing two things, and each deserves its own transaction and its own recorded
reason. Every write path already does exactly one.

**Counting alone is not enough either**, so the matching-move check stays: one
event and one move that describe *different* transitions is still refused.

## Behaviour

| Transaction | Result |
| --- | --- |
| 3 moves, 2 events | refused — *"may change status once per transaction"* |
| 2 moves, 2 events | refused — the rule is one move, not balanced books |
| 1 move, 1 event | accepted |
| 1 move, 1 event describing a different move | refused |
| 2 products moved, 1 event between them | refused |
| An illegal move | still `TP001`, not a provenance complaint |

## Mutation checks

| Mutation | Result |
| --- | --- |
| Revert to the #54 check (*some* event matches) | 2 tests red |
| Keep the count, drop the matching-move check | 1 test red |

## What review got right, and what it did not

**Right:** the 1:1 gap. Real, reproducible, and it would have survived into
every future write path — which is the exact reason #54 was built.

**Already working, verified rather than assumed:** two different products moved
in one transaction with an event for only one of them **was already refused**.
The trigger is row-level, so each product's firing looks for its own event.
There is now a test saying so.

**Not accurate:** the review reports #54 as still open with no pull request, and
the repository at `0` pull requests with lifecycle unbuilt. Checked directly —
#54 is `CLOSED (COMPLETED)` as of the merge of #74. That snapshot is stale.

## Tests

```
apps/web     121 passed      with Postgres: 463 passed
packages/db  282 passed
apps/api      60 passed
```

## Checklist

- [x] The hole reproduced through `psql` before the fix and after it


packages/db/drizzle/0013_one_move_one_event.sql
```diff
@@ -0,0 +1,84 @@
...
+    AND recorded_at >= transaction_timestamp();
+
+  IF explaining_events = 0 THEN
+    RAISE EXCEPTION
+      'A product cannot move from "%" to "%" without recording why, in the same transaction.',
+      OLD.status, NEW.status
+      USING ERRCODE = 'TP004';
+  END IF;
+
+  IF explaining_events > 1 THEN
```
clevervi, line 62: The rule is one move per transaction, and it is turned around deliberately: counting a row trigger's own firings is not possible from inside it, so requiring exactly one event makes the correspondence 1:1 by construction instead of by bookkeeping.

packages/db/drizzle/0013_one_move_one_event.sql
```diff
@@ -0,0 +1,84 @@
...
+
+  IF NOT EXISTS (
+    SELECT 1 FROM lifecycle_event
+    WHERE product_id = NEW.id
+      AND previous_state = OLD.status
+      AND resulting_state = NEW.status
+      AND recorded_at >= transaction_timestamp()
+  ) THEN
+    RAISE EXCEPTION
+      'The event recorded for this product does not describe the move from "%" to "%".',
```
clevervi, line 77: Counting alone is not enough. One event and one move that describe different transitions must still be refused, which is why this check survives alongside the count.

Issue #77, opened 2026-09-18
[CORE] SonarCloud reports 0% coverage, and nothing uploads any


clevervi:
## The misleading signal

SonarCloud shows **0.0% Coverage on New Code** on every pull request. There is no
coverage configuration anywhere in the repository and nothing uploads a report,
so the number is not a measurement — it is the absence of one, displayed as
zero.

A 0% coverage badge on a repository with 463 tests is exactly the class of false
claim this project has spent several pull requests removing from its own
documentation.

## Why it has been left rather than wired up

This project's evidence is **mutation checking**: every guard is broken on
purpose and a named test has to go red. Line coverage does not express that and
can be high while missing it entirely — a suite that executes a guard without
ever asserting it fails counts as covered.

So a coverage number would be a weaker claim sitting next to a stronger one, and
the risk is that it becomes the number people look at.

## What was done instead

`sonar.coverage.exclusions=**/*`, with the reasoning in the file. Nothing claims
a figure nobody produced.

## What would close this properly

Either:

- wire `@vitest/coverage-v8`, upload the report, and treat the figure as
  context rather than a gate — it would at least be true; or
- decide coverage is not a metric this project reports, and say so in
  `CONTRIBUTING.md` beside the mutation-checking bar, so a reader knows the
  absence is a choice.

The second is probably right. The first is worth doing only if the number will
be acted on.

clevervi:
## Correction, from review

This issue over-stated the case. It reads as though mutation checking replaces
coverage. It does not — they measure different things:

| | Answers |
| --- | --- |
| Mutation checking | are the tests sensitive to a change in meaning? |
| Line coverage | what code was executed at all? |

A suite can have an excellent mutation score and still **never execute** whole
paths — error handling, observability, configuration branches. Mutation checking
says nothing about code no test reaches, because there is nothing there to
mutate that anyone would notice.

So the choice is not between them. The honest arrangement:

```
coverage    -> informative, says what was never run
mutation    -> the gate for guards and invariants
integration -> behaviour against real Postgres
```

## What that changes about closing this

The second option originally listed — "decide coverage is not a metric this
project reports" — is the weaker answer. Better:

- wire `@vitest/coverage-v8` and upload the report
- treat the figure as **information**, not a gate, and say so in
  `CONTRIBUTING.md` beside the mutation-checking bar
- keep `sonar.coverage.exclusions` off once a real report exists

The value is not the percentage. It is the list of files at zero, which is the
[...]

PR #78, opened 2026-09-18
fix: require the events to describe the path, not to be exactly one

clevervi:
Authored by [@raishark](https://github.com/raishark). Raised in review of #76,
and the objection was right.

## What #76 got wrong

It required **exactly one** transition event per product per transaction. That
made the correspondence 1:1 — and it did so by forbidding something that was
never wrong:

```
BEGIN;
  UPDATE ... 'suspended'                    -- move 1
  UPDATE ... 'registered'                   -- move 2
  INSERT event registered -> suspended      -- explains move 1
  INSERT event suspended -> registered      -- explains move 2
COMMIT;                                     -- refused, wrongly
```

A transaction deliberately moving a product twice, with a recorded reason for
each move, has nothing to hide. The property that matters is not *one move* but
**the events account for every move**.

Worth noting: the review assumed #76 already allowed this — *"por lo que
describes, sí lo permite"*. It did not. The matrix below is how that was
settled rather than argued.

## The rule now

Counting a row trigger's own firings is impossible from inside it. So instead:

> The transition events written in this transaction, read in order, form an
> **unbroken chain**, and that chain **ends where the product now stands**.

Any single firing can check that, and it is **stronger than counting** — it also
refuses events describing a path the product never took, which a count cannot
see. Each firing additionally asserts its own move appears in the chain, so a
path that is internally consistent and lands correctly still cannot omit a step.

## The matrix, run against real commits

| moves | events | result |
| --- | --- | --- |
| 1 | 0 | FAIL |
| 1 | 1 | **PASS** |
| 1 | 2 | FAIL |
| 2 | 0 | FAIL |
| 2 | 1 | FAIL |
| 2 | 2 | **PASS** |
| 2 | 3 | FAIL |
| 3 | 2 | FAIL |
| 3 | 3 | **PASS** |
| 4 | 4 | **PASS** |

Exactly the matrix review asked for: `moves = events` passes, any imbalance
fails.

## A bug this found, in the first version of this very migration

The chain was compared against `NEW.status`. **The trigger fires once per
`UPDATE`**, and each firing carries the `NEW` from its own statement — so in a
transaction with two moves, the first firing's `NEW` is the *intermediate*
state. Comparing the chain to that refused a perfectly recorded pair of moves.

It now reads the product's **settled** status instead. Only a live run showed
it; the logic reads correctly either way.

## A test-harness lesson worth recording

The first matrix run reported `2 moves + 2 events → FAIL` **after** the fix was
correct. The probe used `SET CONSTRAINTS ALL IMMEDIATE` inside a PL/pgSQL
function, which does not fire deferred triggers the way a real `COMMIT` does.

The harness was wrong, not the rule. Re-run with one real transaction per case,
the matrix came out right — **a reminder that a test which cannot reproduce
production timing is measuring something else.**

## Mutation check

| Mutation | Result |
| --- | --- |
| Revert to #76's one-move-per-transaction rule | 2 tests red |

## Two review points adopted as written

**The invariant is now stated explicitly** in the migration, in the words the
review used: the events must account for every move, and no event may satisfy
two transitions.

**Coverage and mutation testing measure different things.** #77 over-stated the
case for dropping coverage — a suite can have a good mutation score and still
never execute an error path. That issue is updated rather than argued with.

## Tests

```
apps/web     121 passed      with Postgres: 465 passed
packages/db  284 passed
apps/api      60 passed
```

## Checklist

- [x] The matrix run against real commits, not a harness approximation


packages/db/drizzle/0014_events_describe_the_path.sql
```diff
@@ -0,0 +1,116 @@
...
+  END IF;
+
+  -- Against where the product actually ended up, not against NEW.
+  --
+  -- The trigger fires once per UPDATE and each firing carries the NEW from its
+  -- own statement, so in a transaction with two moves the first firing's NEW is
+  -- the intermediate state. Comparing the chain to that would refuse a
+  -- perfectly recorded pair of moves — which is exactly what it did before this
+  -- line was written, and only a live run showed it.
+  SELECT status INTO settled FROM product WHERE id = NEW.id;
```
clevervi, line 91: Against the settled status, not NEW. The trigger fires once per UPDATE and each firing carries its own statement's NEW, so with two moves the first firing sees the intermediate state — which refused a perfectly recorded pair. Only a live run showed it.

packages/db/drizzle/0014_events_describe_the_path.sql
```diff
@@ -0,0 +1,116 @@
...
+  IF chain.steps = 0 THEN
+    RAISE EXCEPTION
+      'A product cannot move from "%" to "%" without recording why, in the same transaction.',
+      OLD.status, NEW.status
+      USING ERRCODE = 'TP004';
+  END IF;
+
+  IF chain.steps > 1 AND NOT coalesce(chain.unbroken, true) THEN
+    RAISE EXCEPTION
+      'The events recorded for this product do not form an unbroken path; each must start where the previous one ended.'
```
clevervi, line 80: Stronger than counting: this refuses events that describe a path the product never took, which a count cannot see. The per-firing check below then stops a consistent chain from omitting a step.

Issue #80, opened 2026-09-18
[IDENTITY] TP-140: ADR — who holds a capacity, before anything authenticates


clevervi:
## Objective

Write the ADR that decides what an identity **is** in TrustPass, before `TP-141` builds a way to present one.

## Why this comes first

`packages/db/src/domain/recording-authority.ts` already answers one half of the question: **what a capacity may assert**. A holder cannot suspend, an issuer cannot reinstate, only the system activates, and only an authority may retire a suspended product.

The other half is unanswered: **who holds that capacity**. `lifecycle_event.actor_kind` is self-declared. Anyone reaching a write path names the capacity they are acting in and the database believes them, which is why `docs/ROADMAP.md:52` says the API must not be exposed publicly yet.

The tempting next step is a users table, a password column and a role enum. That is the wrong shape, and it is expensive to reverse once events reference it. An event is permanent (ADR 0008) — it will point at whatever identity model exists on the day it is written, and a model that cannot express "this authority's warrant expired in March" will have recorded events that cannot be re-read honestly.

Login is the mechanism for **presenting** an identity. It is not the identity.

## What the ADR must decide

For each, the decision and the option rejected:

- **Actor** — the thing that acts. Its relationship to a human being, to an organisation, and to a machine caller.
- **Organisation** — and whether `issuer` is one of these or a separate concept it happens to overlap with.
- **Capacity** — the link back to `RECORDING_AUTHORITY`. Whether an actor holds one capacity or several, and whether it is scoped (all products, or one issuer's).
- **Granting** — who granted the capacity, when, and on what evidence. A capacity nobody granted is a capacity nobody can be asked about.
- **Validity** — valid from, valid until. An authority's warrant is not permanent and a model without an end date cannot say so.
- **Revocation** — how a capacity ends early, and what happens to events recorded under it. They do not disappear: ADR 0008 says the past does not change. So the ADR must say how a passport reads an event recorded by a capacity later revoked.
- **Credential** — what is presented to prove the actor is who they claim. Explicitly deferred to `TP-141`; named here only so the boundary is written down.

## Acceptance criteria

- [ ] `docs/adr/0009-*.md` written in the repository's ADR format, decided rather than surveyed.
- [ ] Every section above answered, each with the alternative that was rejected and what it would have cost.
- [ ] States what it does **not** decide, so `TP-141` has a scope rather than an invitation.
- [ ] Says what happens to events recorded under a capacity later revoked, consistent with ADR 0008.
[...]

PR #81, opened 2026-09-18
security: the database writes recorded_at, not the caller

### LINKED ISSUE #79: [SECURITY] TP-142: a caller-supplied recorded_at defeats the provenance guarantee
## Objective

Make `lifecycle_event.recorded_at` written by the database, not by whoever inserts the row.

## The defect

`0010_provenance_is_guaranteed` and `0014_events_describe_the_path` identify the events belonging to the current transaction with:

```sql
WHERE recorded_at >= transaction_timestamp()
```

`recorded_at` has `defaultNow()` and its doc comment claims it is "set by the database, not by the caller". Neither is true: a default is what happens when nobody supplies a value, and any writer may supply one.

So the trigger decides what "this transaction" means by reading a column the writer chose. That is self-defeating, because these rules live in the database precisely so that a path nobody has written yet cannot bypass them.

## Measured, not assumed

Against the local Postgres, with every trigger enabled:

```
SETUP:    product registered honestly
ATTACK 1: plant an event registered -> suspended, alone in its own
          transaction, with recorded_at = now() + 10 years   -> accepted
ATTACK 2: a LATER, SEPARATE transaction runs
          UPDATE product SET status='suspended'
          and writes no event at all                         -> COMMIT succeeded
          product_now = suspended
```

A status change with no reason recorded in its transaction. That is the exact failure `0010` exists to make impossible, and the history is append-only, so it can never be explained afterwards.

A second consequence, same root cause: `lifecycle_event_not_in_future` is `occurred_at <= recorded_at`. A caller-supplied future `recorded_at` lets a future `occurred_at` through, and an event that has not happened yet now sits in the history table:

```
 type              | occurs_in_the_future | recorded_at
-------------------+----------------------+---------------------
 record_corrected  | t                    | 2032-09-18 ...
```

Third consequence: `apps/api/src/passports/read-passport.ts:143` publishes `recordedOn` on the public passport. Today that is a caller-controlled string presented as when TrustPass learned of something.

## Acceptance criteria

- [ ] A `BEFORE INSERT` trigger on `lifecycle_event` overwrites `recorded_at` with `now()`, for every writer including the owner, as `lifecycle_event_append_only` already does for UPDATE and DELETE.
- [ ] A supplied `recorded_at` is ignored rather than rejected. A `BEFORE` trigger cannot distinguish a caller passing `now()` from the default firing, so rejection is not implementable; overwriting is.
- [ ] `NewLifecycleEvent` no longer accepts `recordedAt`, so the compiler refuses the attempt before Postgres has to.
- [ ] Integration test: the planted-event sequence above is refused with `TP004`.
- [ ] Integration test: a supplied `recorded_at` is not the stored value.
- [ ] Integration test: a future `occurred_at` can no longer ride in on a future `recorded_at`.
- [ ] Every one of those mutation-checked: remove the trigger, watch each named test go red.
- [ ] The write-order protocol documented where the table is defined (see below).

## Also in scope: the event write order is a protocol, not an accident

`0014` orders the chain with `ORDER BY id`. Raised in review, and correct: that makes the order in which events are written part of the meaning of the history. Writing the event for the second move before the event for the first produces a chain that does not match the path, and the transaction is refused.

That is the right behaviour — an event written before its cause is not a record of anything — but today it is a property of an identity column rather than a decision. Document it as a rule at the table:

> Transition events are written in the same order as the transitions they describe.

## Risk

`critical`. `CONTRIBUTING.md` defines it as "a mistake here corrupts the trust model itself", and a status change with no recorded reason is exactly that.

## Dependencies

None. Forward-only migration, additive.

clevervi:
## What changed

`lifecycle_event.recorded_at` is now overwritten by a `BEFORE INSERT` trigger, for every writer including the owner. It had `defaultNow()` and a doc comment claiming it was "set by the database, not by the caller" — but a default is what happens when nobody supplies a value, and the provenance triggers decide what "this transaction" means by reading precisely this column. The guard was reading a value the writer chose.

## Why

Closes #79

## The defect, measured rather than argued

Run against the local Postgres with every trigger enabled, **before** the fix:

```
SETUP:    a product registered honestly
ATTACK 1: plant an event registered -> suspended, alone in its own transaction,
          with recorded_at = now() + 10 years            -> accepted
ATTACK 2: a LATER, SEPARATE transaction runs
          UPDATE product SET status='suspended'
          and writes no event at all                     -> COMMIT succeeded
          product_now = suspended
```

A status change with no reason recorded in its transaction — the one thing `0010_provenance_is_guaranteed` exists to make impossible — on a history that is append-only and therefore could never be corrected afterwards.

**After** the fix, the identical script:

```
ATTACK 2: ERROR: A product cannot move from "registered" to "suspended"
                 without recording why, in the same transaction.   [TP004]
          product_now = registered
```

Two further consequences of the same column, both closed by the same trigger:

- `lifecycle_event_not_in_future` is `occurred_at <= recorded_at`, so a caller-supplied future `recorded_at` carried a future `occurred_at` past it. An event that has not happened yet is not an event (ADR 0008), and a passport would have rendered it as a fact. The failing row after the fix shows the trigger doing its work first: `occurred_at 2031-…, recorded_at 2026-…` → `violates check constraint "lifecycle_event_not_in_future"`.
- `apps/api/src/passports/read-passport.ts` publishes `recordedOn` on the public passport. Until now that was a caller-controlled value shown to a reader as when TrustPass learned of something.

## Acceptance criteria

- [x] `BEFORE INSERT` trigger overwrites `recorded_at` with `now()`, for every writer including the owner
- [x] Supplied values ignored rather than rejected — see "the choice that was forced" below
- [x] `NewLifecycleEvent` no longer accepts `recordedAt`
- [x] Integration test: the planted-event sequence is refused with `TP004`
- [x] Integration test: a supplied `recorded_at` is not the stored value
- [x] Integration test: a future `occurred_at` can no longer ride in on a future `recorded_at`
- [x] All three mutation-checked
- [x] The write-order protocol documented at the table

## The choice that was forced, not preferred

Overwriting is not a softer option chosen over rejecting. **Rejecting is not implementable here.** The column default fills `recorded_at` in before any `BEFORE` trigger runs, so by the time the function sees `NEW` there is no way to distinguish a caller who passed `now()` from a caller who passed nothing. A guard that cannot tell two cases apart cannot refuse one of them. Assignment reaches the same end state without claiming a detection it does not have.

`now()` rather than `clock_timestamp()`: `now()` *is* `transaction_timestamp()`, so every event written in a transaction lands on exactly the value `0014`'s filter compares against, and every event from an earlier transaction falls below it.

## Also: the event write order is a protocol now, not an accident

`0014` orders the chain with `ORDER BY id`. Raised in review, and the objection was right: that makes the order events are written in part of the meaning of the history. Writing the second move's event before the first move's produces a chain that does not match the path, and the transaction is refused.

That refusal is wanted — an event written before its cause records nothing — but it was a property of an identity column rather than a decision anybody had recorded. It is now written at the table, so the next person to touch the ordering knows they are changing what the history means.

## Tests

3 added, 468 total (121 web, 287 db, 60 api). All four gates green locally.

### Mutation table

The trigger dropped, the suite re-run, the trigger restored:

| Guard broken | Test that went red | How it failed |
| --- | --- | --- |
| `lifecycle_event_recorded_at_is_ours` | `ignores a recorded_at the caller supplies` | `expected false to be true` — the stored value was ten years ahead of `now()` |
| same | `no longer lets a future occurrence ride in on a future recorded_at` | `Expected SQLSTATE 23514, but the operation succeeded` |
| same | `does not accept an event planted with a future recorded_at` | `Expected SQLSTATE TP004, but the operation succeeded` |

```
Tests  3 failed | 284 passed (287)
```

Exactly the three, and nothing else — which is the second half of the check. A guard that breaks unrelated tests is either too broad or the tests are measuring it by accident.

The first test compares `recorded_at <= now()` **in SQL**, not in JavaScript. The host clock runs ahead of the container's, so a `Date` taken in the test is not the same instant as `now()`, and a comparison between them measures the skew rather than the guarantee. That mistake already cost a day under TP-051.

## Security impact

This **is** the security impact. It closes a path by which any SQL writer could move a product's status with no recorded reason, and could place an event that has not happened into an append-only history.

No new dependency, no new external call, no change to auth (there is none yet — that is #80 and `TP-141`). No API shape change.

## Breaking changes

Schema: additive, forward-only, one new trigger and one new function. `0015_the_database_writes_recorded_at`.

Type: `NewLifecycleEvent` no longer accepts `recordedAt`. Nothing in the repository supplied it, so nothing broke; a future caller now gets a compile error instead of a silently discarded value.

---

- [ ] CI is green
- [x] Documentation updated: the column, the table's write-order protocol, and the migration's own reasoning


clevervi:

Labelled per `CONTRIBUTING.md`. One person wrote this and one person is reading it; that is the same judgement running twice, and the only thing that makes it worth anything is that the reasoning is written where someone else can check it.

### A weakness I found in my own test, and fixed

`recorded_at <= now()` also accepts `1970-01-01`. It would have passed for a trigger that wrote **any** past constant — a different defect wearing this one's clothes. The requirement is that the column holds the time of the write, so the assertion now says that, bounded from both sides with a minute of room for the host-container clock skew. Pushed as `27d3ab1`.

### A mutation result I did not earn, reported as such

Two mutations were run. They are not equally conclusive and the difference matters.

**Mutation A — drop the trigger.** Conclusive:

```
Tests  3 failed | 284 passed (287)
```

Exactly the three new tests, each failing on its own named reason (`expected false to be true`, `Expected SQLSTATE 23514`, `Expected SQLSTATE TP004`), and **nothing else moved**. That second half matters as much as the first: a guard whose removal breaks unrelated tests is either too broad or is being measured by accident.

**Mutation B — trigger writes `timestamptz '1970-01-01'` instead of `now()`.** Caught, but **not by the test I tightened**:

```
Tests  101 failed | 134 passed | 52 skipped (287)
```

Every provenance fixture dies, because `insertProductWithProvenance` can no longer write an event this transaction can see. Which means the `beforeAll` of `lifecycle-event.integration.test.ts` dies first, and `ignores a recorded_at the caller supplies` **never runs** — verified by running that test alone under the mutation: `25 skipped`.

So the past-constant mutation is caught by the suite in aggregate, not by the test that names the property. The tightening is still right — it states the actual requirement rather than a weaker one that happens to hold — but I am not going to write it up as a mutation the test caught, because it did not.

### Two things worth disagreeing with

**1. `packages/db/src/schema/lifecycle-event.integration.test.ts` — the `planted()` helper casts through `as NewLifecycleEvent`.**

That is a deliberate lie to the compiler, in a repository that just spent a commit making the compiler refuse this exact field. The argument for it: the type is not the guard, it only stops an honest mistake, and the claim under test is that the *database* refuses a supplied value. A test that could not express the illegal call could not test the guard at all. Still, it is a cast, and casts are how a type system stops meaning anything. Worth a second opinion.

**2. `0015` argues `now()` over `clock_timestamp()` on the strength of `0014`'s `>=` filter.**

That reasoning is correct today and has a short shelf life: #82 proposes replacing the timestamp comparison with `pg_current_xact_id()` equality, which removes the premise. The migration is forward-only and will not be edited, so this note is the record that its justification was scoped to the rule in force when it was written.

### What this does not fix, found while verifying it

#82, opened during this work and reproduced against **this branch**, with `0015` applied:

```
A: BEGIN                                    (transaction_timestamp = T0)
B: BEGIN at T1 > T0, INSERT event reg -> susp, COMMIT
A: UPDATE product SET status='suspended'    (writes no event at all)
A: COMMIT                            -> ACCEPTED
```

`recorded_at >= transaction_timestamp()` is a half-open **time window**, not an identity, and under READ COMMITTED another transaction's committed event falls inside it.

Deliberately out of scope here. It is a different root cause needing a schema change, and #79 is not superseded by it — any fix depends on the column being server-written, which is what this PR establishes. **Nobody should read this PR as making provenance airtight.** It closes one of two doors.

PR #83, opened 2026-09-18
docs: decide what an identity is, before building one to present

### LINKED ISSUE #80: [IDENTITY] TP-140: ADR — who holds a capacity, before anything authenticates

clevervi:
## What changed

ADR 0009 — what an identity **is** in TrustPass. No code. Registered in the ADR index, and `docs/ROADMAP.md` now says `TP-141` is gated on this rather than the reverse.

## Why

Closes #80

`recording-authority.ts` already decides **what a capacity may assert**, and a trigger enforces it: a holder cannot reinstate, an issuer cannot clear a suspension, only an authority may retire a suspended product. Nothing decides **who holds one**. `actor_kind` is self-declared and the database believes whoever writes it, which is why the roadmap says the API must not be exposed publicly yet.

The tempting next step is a `user` table, a password column and a `role` enum. It ships this week and fails on four questions the system already needs answered:

| Question | A role column's answer |
| --- | --- |
| Who granted this capacity? | Nothing. It was set by an UPDATE. |
| On what evidence? | Nothing. |
| Until when? | Never. A role has no end. |
| What happens to what they recorded, after it is taken away? | The column changes and the history silently re-reads. |

Rejected on the fourth, not the first three. Flipping `role` from `authority` to `null` retroactively changes what every past event means, in a table whose entire purpose is that the past does not change.

## What it decides

Nine decisions, each with the alternative rejected and what it would have cost. The three worth arguing about:

**An event records the grant it acted under, not only the capacity.** Reads like an extra column. Without it, revoking an authority's warrant leaves every event it wrote indistinguishable from events written by an authority in good standing — unfindable, and therefore unqualifiable.

**Revocation does not invalidate the past.** An event records that at time T an actor holding a then-valid capacity recorded something, and that stays true; what changes is how much it is worth. So the passport gains a qualifier — *"recorded by an authority whose warrant was later withdrawn"* — never an erasure. Deleting or hiding those events would be `0005_lifecycle_event_append_only`'s failure arriving through the identity model instead of through SQL. Treating them as still fully authoritative is the mirror-image dishonesty.

**A holder holds a capacity nobody granted.** No registry of people who own things exists, and inventing a grant for one would fabricate an authority that does not. So it is recorded as self-asserted rather than papered over — ADR 0007 already says a holder-origin record establishes only that a serial was entered.

## Measured while writing it

The claim that an `authority` event names no authority is stronger than "the column is missing". The schema **requires** the omission:

```sql
check("lifecycle_event_issuer_matches_actor",
      (actor_kind = 'issuer') = (issuer_id IS NOT NULL))
```

An authority event must have `issuer_id IS NULL`. There is no column in which to name the authority, so a passport can report a theft and structurally cannot say who reported it. The constraint is right for what it guards — a half-recorded actor is worse than none — but the party it could point at does not exist yet.

That is what pushes decision §2: an organisation is the party, and issuer is a role it plays.

## What it deliberately does not decide

So `TP-141` has a scope rather than an invitation: the credential mechanism, sessions, token lifetime, how an organisation gets verified in the first place, whether grants are delegable, the public wording for a revoked grant, rate limiting.

### For #49

The ADR supplies the asymmetry that makes the case decidable — a manufacturer's claim comes from a granted, scoped, evidenced capacity and a holder's is self-asserted by construction. Those are not two equal claims about one object. It does **not** decide what happens to the two records; merge, link or supersede is still #49's work.

## Tests

None. No code changed.

The honest note: an ADR is not testable, and this one is load-bearing — every event written after `TP-141` will point at this model forever, because events are permanent (ADR 0008). There is no migration back to a fact nobody wrote down. **This is the artefact in the repository most worth disagreeing with**, and it is worth more disagreement than the PRs that have tests.

## Security impact

Decides the shape of the authorisation model. Nothing is enforced by this PR; `actor_kind` remains self-declared until `TP-141`, and the roadmap's warning about not exposing the API publicly still stands.

Names a new risk explicitly: a grant becomes the thing worth stealing, so `TP-101` and `TP-104` stop being generic hardening.

## Breaking changes

None now. States a direction that is expensive later: `issuer` becomes one role of an `organisation`, which means rewriting every foreign key pointing at an issuer. Decided now precisely because there will be more of them each week.

---

- [ ] CI is green
- [x] `pnpm lint` passes locally; typecheck, test and build unaffected — no code changed


clevervi:

Labelled per `CONTRIBUTING.md`. One person wrote this and one person is reading it. The mutation bar does not apply — there is no code, and saying "N/A" is more useful than inventing a check.

An ADR is the artefact where a self-review is worth least, because there is no suite to disagree with me. So here are the three places I think this is weakest, stated as plainly as I can manage.

### 1. §2 is the most expensive decision in the document and has the weakest forcing argument

Making `issuer` a role of an `organisation` means rewriting every foreign key that points at an issuer. That is the biggest bill in here.

The gap it is justified by — an `authority` event cannot name the authority — **does not require it**. Adding an `authority` table and an `authority_id` column would close that gap at a fraction of the cost.

My argument against that is the one ADR 0006 makes: one real-world entity should not exist twice with two verification states that can disagree. A manufacturer that also runs its own service centre would be both an issuer and a repairer.

**That manufacturer does not exist in this system.** There is no repairer, no insurer and no authority record of any kind. I am generalising on a predicted need, and "we might need it later" is the reasoning this repository spent a day deleting from the roadmap.

I still think it is right, because the cost of the narrow fix is paid again per kind and this schema is going to grow more organisation-like parties, not fewer. But it is a judgement call dressed up as a consequence, and if one decision here gets overturned it should be this one.

### 2. §6 puts a column on an append-only table for a benefit that may never arrive

Recording the grant on every event costs storage on every row forever, and pays only when a grant is revoked. If no warrant is ever withdrawn, it is dead weight.

The counter is the real one — it cannot be added retroactively, because the events that needed it were written before it existed — but that argument justifies almost any column, so it deserves less confidence than its shape suggests.

### 3. "There is no migration back to a fact nobody wrote down" proves too much

It is true, and taken alone it argues for capturing everything now. It does not, by itself, justify the particular boundary I drew — grant, scope, evidence, validity, revocation in; jurisdiction, delegation depth, and the verification process out.

The boundary is a judgement about which facts are likely enough to be needed, not a consequence of the principle. I did not defend it in the document and I am naming that here rather than leaving it to read as rigour.

### What I am confident about

§7 — revocation does not invalidate past events, and the passport gains a qualifier rather than an erasure. That one is not a preference. A history that edits itself when somebody's standing changes is not a history, and both failure modes are dishonest in opposite directions: deleting them hides what was reported, and leaving them unqualified hides that the reporter's warrant was withdrawn.

§8 — a holder capacity is granted by nobody and is recorded as self-asserted. Papering over that would be the first outright lie in the model.

PR #84, opened 2026-09-18
security: identify this transaction, do not date it

### LINKED ISSUE #82: [SECURITY] TP-143: "this transaction" is a time window, so a concurrent transaction's event explains your move
## Objective

Identify the events belonging to the current transaction by **the transaction**, not by a timestamp range.

## The defect

`0010` and `0014` scope the provenance chain with:

```sql
WHERE recorded_at >= transaction_timestamp()
```

That is a half-open time window, not an identity. Any event another transaction commits **after this one began** falls inside it. Under READ COMMITTED — Postgres's default and what this application uses — a later statement in transaction A sees rows transaction B committed in the meantime.

So A's status change can be explained by B's event.

## Measured, not assumed

Two connections against the local Postgres, every trigger enabled, **including the `0015` fix from #79**:

```
setup:  product registered honestly
A:      BEGIN                                     (transaction_timestamp = T0)
B:      BEGIN at T1 > T0
        INSERT event registered -> suspended
        COMMIT
A:      UPDATE product SET status='suspended'     (writes no event at all)
A:      COMMIT                              -> ACCEPTED
product now: suspended
```

A moved the product having recorded nothing. B's event, written by a different transaction for a different reason, accounted for it.

## Why #79 does not cover this, and why it was still necessary

Different root cause. #79 was "the caller chooses the value the guard reads". This is "the value the guard reads is a time range, and a range has other people in it".

#79 remains necessary rather than superseded: **any** scheme here depends on the column being written by the server. If `recorded_in_xact` were addable by a caller, this rule would be bypassable exactly as `recorded_at` was. #79 establishes the principle; this issue changes what is recorded.

## Proposed direction

`pg_current_xact_id()` returns `xid8` — 64-bit, does not wrap, available since PG13; this runs PG18. A column defaulting to it, **forced by the same `BEFORE INSERT` trigger #79 introduced**, turns the filter into an equality:

```sql
WHERE recorded_in_xact = pg_current_xact_id()
```

No window, no concurrency, no clock. The event either was written by this transaction or it was not.

Costs to weigh in the PR, not here:

- A new column on an append-only table, so it is additive and forward-only, but it is 8 bytes on every historical row.
- `pg_current_xact_id()` **assigns** a transaction id if the transaction does not have one yet. A read-only transaction that touches `lifecycle_event` would start consuming xids. Whether that matters at this volume needs measuring rather than guessing.
- Whether `recorded_at` stays. It should: it answers "when did TrustPass learn of this", which is a different question from "which write produced this" and is what the passport publishes.

## Acceptance criteria

- [ ] The provenance triggers identify this transaction's events by transaction identity, not by a timestamp comparison.
- [ ] The column is server-written, per #79.
- [ ] Integration test reproducing the two-connection sequence above, refused with `TP004`.
- [ ] That test mutation-checked: it must go red when the change is reverted.
- [ ] The xid-assignment cost measured against a read path and written in the PR, not asserted.
- [ ] `0015`'s reasoning updated where it now reads as complete.

## Risk

`critical`. Same reason as #79: a status change with no recorded reason, on an append-only history that can never be corrected.

## Dependencies

#79 / #81 first. This builds on the trigger that one introduces.

clevervi:
## What changed

The provenance triggers stop asking *when* an event was written and start asking *which write produced it*. `lifecycle_event` gains `recorded_in_xact xid8`, set by the trigger `0015` introduced, and the chain is scoped by `recorded_in_xact = pg_current_xact_id()`.

## Why

Closes #82

`0010` and `0014` scoped the chain with `recorded_at >= transaction_timestamp()`. That is a half-open **range**, not an identity, and it answers a different question than the one intended:

```
intended:  did THIS transaction write this event?
asked:     was this event written after this transaction started?
```

Under READ COMMITTED — Postgres's default and what this application runs on — a later statement sees rows other transactions committed in the meantime. So the answer was *yes* for events this transaction had nothing to do with.

## Measured, with two connections

**Before**, against a database with `0015` already applied:

```
A  BEGIN, one statement           (transaction_timestamp = T0)
B      BEGIN at T1 > T0
B      INSERT event registered -> suspended
B      COMMIT
A  UPDATE product SET status='suspended'      -- writes no event at all
A  COMMIT                                     -> ACCEPTED
```

**After**, the identical script:

```
A moved the product with no event of its own:
  REFUSED  TP004  A product cannot move from "registered" to "suspended"
                  without recording why, in the same transaction.
product now: registered
```

### The range was not only too permissive — it also refused correct work

Found while mutating, and not something the issue predicted. Under the time window, A writing **its own** event while B commits one puts *both* in the chain, they do not connect, and the transaction is refused. A range does not merely let the wrong thing in; it breaks a transaction that did everything right. Pinned by the second test below.

## Verified before relying on it, not read from the docs

Every claim the fix rests on, run against this database:

| Claim | Result |
| --- | --- |
| `pg_current_xact_id()` is the **top-level** id, from a savepoint | `top level 14601` / `inside savepoint 14601` / `nested savepoint 14601` — identical |
| A plain `SELECT` assigns no transaction id | `pg_current_xact_id_if_assigned() IS NULL` → true |
| Calling `pg_current_xact_id()` *does* assign one | NULL before, non-NULL after |
| `'0'::xid8` can never equal a live id | Postgres reserves 0–2 and hands out nothing below 3 |
| A transaction's timestamp is fixed by its **first statement**, not by `BEGIN` | 0.4s `pg_sleep` before reading it → the gap is real |

The last one is why the test harness runs a statement before letting B commit. Without it A's timestamp lands *after* B's write and the ordering the test is named for does not exist — it would have passed for the wrong reason.

## The XID cost, measured rather than waved away

`pg_current_xact_id()` assigns an id if the transaction has none, which is the documented reason `pg_current_xact_id_if_assigned()` exists. The concern is read paths starting to consume ids.

They do not, and here is the measurement rather than the argument:

```
1. the real passport read path, inside a transaction
   (product LEFT JOIN issuer, then its lifecycle_event history)
   -> read_path_assigned_no_xid = t

2. 200 passport-style reads
   before_xmax 16662 | after_xmax 16662 | xids_consumed = 0
```

Structurally: the only caller is a `BEFORE INSERT` trigger, and an `INSERT` has already forced an id. No `SELECT` fires it. So the assigning form costs nothing here, while `_if_assigned` could return NULL into a `NOT NULL` column.

## The `'0'` default is load-bearing

Two things at once:

1. Rows written before the column existed carry `'0'` — "written by a transaction this database can no longer name". True, and better than backfilling this migration's own id, which would claim every historical event came from one statement.
2. **It fails closed.** Drop the trigger and new rows take `'0'`, which matches no transaction, so every provenance check refuses loudly. Compare `recorded_at`: `defaultNow()` satisfied the old filter, so losing *that* trigger would have failed **open**, in silence. Proven by mutation 2 below.

## `recorded_at` stays

Two different facts, and one column doing both jobs is exactly what #79 was:

```
recorded_at       = when did TrustPass learn of this   (the passport publishes it)
recorded_in_xact  = which write produced this row      (nothing outside ever sees it)
```

## Tests

6 added, 474 total (121 web, 293 db, 60 api). All four gates green locally.

### Mutation table

**Mutation 1 — restore `recorded_at >= transaction_timestamp()`** (re-run `0014`'s function, putting the defect back):

| Test that went red | Why |
| --- | --- |
| `refuses A's move when only B recorded an event for it` | The move was accepted |
| `accepts A's move when A recorded its own event, even while B commits one too` | Both events fell in the window and did not connect — a correct transaction refused |
| `does not let B's event break a chain A recorded correctly` | B's row entered A's chain |

```
Tests  3 failed | 290 passed (293)
```

**Mutation 2 — trigger stops setting `recorded_in_xact`**, so the `'0'` default takes over:

```
Tests  53 failed | 240 passed (293)
```

Every provenance path refuses, loudly, with `TP004 ... without recording why`. That is the fail-closed property, demonstrated rather than claimed.

### One test was wrong, and mutating is what found it

`does not let B's event break a chain A recorded correctly` originally had B write `suspended -> registered`. That chains cleanly in front of A's two events, so it **passed under the time window too** — green for a reason it did not name, which is the exact failure `CONTRIBUTING.md` warns about. B now duplicates A's first move, so the broken chain is the one B causes. It went red under mutation 1 only after that change.

### Two tests no mutation here kills, said plainly

The savepoint pair pins that the unit of provenance is the commit — an event inside a nested transaction counts, an event rolled back to a savepoint does not. **Neither mutation kills them**, because both mutations leave that behaviour intact, and there is no natural mutation for "return a subtransaction id" since Postgres exposes no such function. They are characterisation tests: they document behaviour the fix depends on and would catch a future change that broke it. Calling them mutation-verified would be inflating the score, which is worse than a low one.

## Security impact

This **is** the security impact. It closes the second of the two doors #79 left; the first was closed by #81. A provenance guarantee that another transaction could satisfy was not a guarantee.

No new dependency, no new external call, no API shape change, no change to auth (still none — #80 / `TP-141`).

## Breaking changes

Schema: additive and forward-only. One column, one function extended, one function replaced. `0016_this_transaction_is_an_identity`.

Type: `NewLifecycleEvent` also omits `recordedInXact`. Nothing supplied it.

Storage: 8 bytes per row on an append-only table that will only grow. A conscious cost, recorded here rather than discovered later. Partitioning or compressing the history is a real question at a size this project is nowhere near, and not one to pre-solve.

---

- [ ] CI is green
- [x] Documentation updated: the column, the migration's reasoning, and the harness


clevervi:

Labelled per `CONTRIBUTING.md`. One person wrote this and one person is reading it; the only thing that makes it worth anything is that the reasoning is somewhere a third party can check it.

Two accounts authored this branch and that is authorship, not review. `1b67511` and `73019b8` are Raishark's, `793588e` is clevervi's. Neither approved the other's work and the merge will not pretend otherwise.

### What mutating found that reading did not

Twice now, in one branch.

**A test of mine passed for a reason it did not name.** `does not let B's event break a chain A recorded correctly` had B write `suspended -> registered`. That happens to chain cleanly in front of A's two events, so it stayed green *under the defect as well*. I only found it because mutation 1 reported 2 failures where I had predicted 3, and the missing one was the test I would have sworn was the most direct. Fixed by having B duplicate A's first move; it goes red now.

**The defect was worse than #82 described.** The issue framed the time window as too permissive. It is also too *restrictive*: under it, A writing its own event while B commits one puts both in the chain, they do not connect, and a completely correct transaction is refused. I did not predict that and would not have found it by reading — the second test caught it.

**1. A latent problem this branch surfaced but does not fix — #85.**

Checking for schema drift, `pnpm db:generate` wanted to emit:

```sql
ALTER TYPE "public"."lifecycle_event_type" ADD VALUE 'product_activated' …
```

for a value added back in `0011`. `drizzle-kit`'s snapshots only update when *it* generates a migration, and every migration since `0011` has been hand-written, so they describe a database that no longer exists. `ALTER TYPE … ADD VALUE` on an existing value fails, so the next person to run `db:generate` inherits a broken migration nobody wrote.

Deliberately out of scope here — it predates this branch by five migrations and fixing it means deciding how migrations are authored. But it is the kind of thing that is invisible until it bites, so #85 exists rather than a note in a commit message.

**2. `xid8` is declared with `customType`, which is a small bet.**

Drizzle has no built-in for it, so `packages/db/src/schema/lifecycle-event.ts` declares one. If a future Drizzle adds `xid8` natively there will be two definitions and the local one wins silently. Low stakes — the only operation is equality and the database performs it — but it is a private reimplementation of something a dependency may later own.

**3. `recordedInXact` is now in the select type and nothing reads it.**

It is an integrity column; no application code should ever look at it. It appears in `LifecycleEvent` anyway, because `$inferSelect` takes every column. Omitting it from the insert type was possible and done; omitting it from the read type is not, without hand-writing the row type. So the type surface now advertises a field whose correct usage is "never". Worth a second opinion on whether that is worse than the alternative.

### Two tests I am not claiming credit for

The savepoint pair is not killed by either mutation, and the PR body says so. There is no natural mutation for "return a subtransaction id" because Postgres exposes no such function, so those two are characterisation tests: they pin behaviour the fix depends on and would catch a future change that broke it, but no mutation here demonstrates they can fail.

A mutation score that counts them would be inflated, and an inflated one is more dangerous than a low one — it is the thing that stops you looking.

### What I am confident about

The `'0'` default failing closed, which mutation 2 demonstrates rather than asserts: 53 tests red, every one with `TP004 … without recording why`. That is the property `recorded_at` did not have — its `defaultNow()` satisfied the old filter, so losing that trigger would have failed open and in silence.

And the fix is isolation-level independent. Transaction-id equality holds the same under READ COMMITTED, REPEATABLE READ and SERIALIZABLE, where the time window's behaviour was a consequence of READ COMMITTED's visibility rules. The guarantee no longer depends on a setting nobody is watching.

clevervi:
## Addendum — the quality gate, and a guess that was wrong

SonarCloud refused this branch twice before passing, and how it got resolved is worth recording because the first attempt was aimed at the wrong thing.

**First failure: 7.9% duplication on new code**, against a 3% gate. The obvious cause was the concurrency suite having copied the product factory and the transition-event builder out of the single-connection suite. Extracted both to `testing/provenance-fixtures.ts` (`9ea1000`).

**Second failure: 7.5%.** Barely moved. So that was not it.

The next suspect was `0016`'s SQL, which genuinely does restate almost all of `0014`'s function — `CREATE OR REPLACE` has no alternative, and editing an applied migration is forbidden. The instinct was to widen `sonar.cpd.exclusions`.

**That would have been wrong, and asking rather than guessing is what showed it.** Querying Sonar for the per-file numbers instead:

```
 25.4%  new_lines=177  <== packages/db/src/schema/provenance-concurrency.integration.test.ts
  0.0%  new_lines=175      packages/db/drizzle/0016_this_transaction_is_an_identity.sql
```

The migration is at **0.0%**. `sonar.cpd.exclusions=packages/db/drizzle/*.sql` has been working exactly as written. Loosening it would have changed no number and quietly weakened a rule that was doing its job.

The duplication was 45 of the 177 lines of the concurrency suite itself: five tests each spelling out the same status change, the same event insert and the same read-back. `moves`, `records` and `settledStatus` name those three steps (`de9c39a`), and the gate passes.

### Mutation 1 was re-run after **each** refactor

Both times, and this is the part that is not optional. Deduplicating tests is the easiest way to weaken them without noticing — a shared helper that quietly softens an assertion leaves a green suite and a guard nobody is checking any more.

```
after 9ea1000 (fixture extraction):    Tests  3 failed | 290 passed (293)
after de9c39a (step extraction):       Tests  3 failed | 290 passed (293)
```

Same three tests, same reasons. The refactors moved code, not evidence.

### Final state

**474 tests** — 121 web, 293 db, 60 api. All seven checks green.

Four commits, two accounts: `1b67511` and `73019b8` and `de9c39a` are Raishark's, `793588e` and `9ea1000` are clevervi's. That is shared authorship and it is disclosed; neither account reviewed or approved the other's work, and the rebase merge keeps the record of who wrote what rather than collapsing it into whoever pressed the button.

PR #86, opened 2026-09-18
chore: make the migrations describe the schema, and keep them that way

### LINKED ISSUE #85: [CORE] TP-144: drizzle-kit's snapshots have been stale since 0011, so db:generate emits migrations that already ran
## What happens

Running `pnpm db:generate` on a clean, fully-migrated checkout emits a migration for changes that are **already applied**:

```sql
ALTER TYPE "public"."lifecycle_event_type" ADD VALUE 'product_activated' BEFORE 'product_retired';
```

Found while checking for schema drift during #84, and confirmed to predate it: `product_activated` was added by `0011_product_activated_event.sql`, and **no snapshot in `packages/db/drizzle/meta/` contains it**.

## Why

Migrations in this repository are hand-written on purpose — the triggers, the deferred constraints and the partial indexes are things `drizzle-kit` cannot express, and `CONTRIBUTING.md` says migrations are forward-only and never edited.

But `drizzle-kit` keeps its own idea of the schema in `meta/NNNN_snapshot.json`, and it only updates that when **it** generates a migration. A hand-written migration never touches the snapshot, so from `0011` onward the snapshots describe a database that no longer exists.

## Why it matters

Not today's correctness — the database is right and the tests run against it. It is a trap:

- Anyone running `db:generate` gets a migration re-applying old changes. `ALTER TYPE … ADD VALUE` on an existing value **fails**, so the next person's first experience is a broken migration they did not write.
- The drift check that caught this is only available by reading the generated file and throwing it away. There is no signal that says "your snapshot is behind".
- It gets worse per hand-written migration, and every migration since `0011` has been hand-written.

## Options, none chosen yet

- **Regenerate the snapshots to match reality**, once, and accept that they must be refreshed whenever a migration is hand-written. Cheap now, same trap later.
- **Stop using `drizzle-kit generate` entirely** and write every migration by hand, deleting the script so nobody reaches for it. Honest about what this repository actually does; loses the diffing help for ordinary column changes.
- **Keep the schema authoritative and add a CI check** that `db:generate` produces no diff, failing the build when schema and migrations disagree. Catches real drift in both directions, and is the only option that would have caught this in `0011` rather than in `0016`.

The third is the one worth costing. It turns a silent divergence into a red build, which is what this repository does with every other rule it cares about.

## Acceptance criteria

- [ ] `pnpm db:generate` on a fully-migrated checkout produces no migration.
- [ ] Whatever keeps it that way runs in CI, not in someone's memory.
- [ ] The decision recorded where migrations are documented, since it changes how they are written.

## Risk

`medium`. Crosses a module boundary and shapes how every future migration is authored, but nothing in the running system is wrong today.

clevervi:
## What changed

`drizzle-kit`'s snapshot catches up with the schema, and CI fails the build if it ever falls behind again.

## Why

Closes #85

`drizzle-kit` refreshes `meta/NNNN_snapshot.json` **only when drizzle-kit itself generates a migration**. Every migration since `0011` has been hand-written — triggers, deferred constraint triggers and partial indexes are not things the generator can express — so the snapshots described a database that stopped existing five migrations ago, and `pnpm db:generate` on a clean, fully-migrated checkout offered:

```sql
ALTER TYPE "public"."lifecycle_event_type" ADD VALUE 'product_activated' BEFORE 'product_retired';
ALTER TABLE "lifecycle_event" ADD COLUMN "recorded_in_xact" "xid8" DEFAULT '0' NOT NULL;
```

Both **fail** against any database that has run `0011` and `0016` — which is every database, a freshly reset one included, because those migrations create exactly what the generator offered to add. Nothing deployed was wrong; the tool meant to help you write the next migration was lying about the current state.

## Option C, not option A

The issue offered three. This takes the third, and the reasoning is that the other two do not hold.

Regenerating the snapshot once fixes today and nothing else — the next hand-written migration drifts again, silently, and the next person to run `db:generate` inherits a migration nobody wrote. Deleting `db:generate` is honest but throws away real help for ordinary column changes.

So: reconcile **and** check. The check is what makes the reconciliation stay true.

### `--custom` is the obvious tool and is the wrong one

Tried first. It writes an empty migration but copies the **previous** snapshot forward, so the drift survives — measured, the very next `generate` produced an `0018` carrying the same two statements.

What works is letting `generate` write the migration, which produces a correct snapshot, then replacing the SQL body with an explanation of why it is empty. The correct migration here **is** no migration: both statements fail everywhere, so `0017` exists for its snapshot.

## Verified against a database built from zero

Not the one that was already running, because that one could have been carrying the answer:

```
docker volume destroyed
all seventeen migrations applied      -> migrations applied successfully
pnpm db:generate                      -> No schema changes, nothing to migrate
full suite                            -> 474 passed (121 web, 293 db, 60 api)
```

## The check, verified in both directions

A guard that cannot fail is not a guard.

| Case | Result |
| --- | --- |
| Schema and migrations agree | `No schema changes, nothing to migrate` — nothing produced, build passes |
| A column added to `issuer` with no migration written | `0018_wooden_toad.sql` produced, build fails, printing `ALTER TABLE "issuer" ADD COLUMN "drift_probe" varchar(10)` |

### Placed in the existing `quality` job on purpose

It needs no database — `generate` diffs the schema against its snapshot and never opens a connection, verified by running it against `postgresql://unused:unused@127.0.0.1:1/unused`.

And branch protection pins the required checks **by name**:

```
["Lint, typecheck and build", "Test against Postgres", "Secret scan"]
```

A new job would be advisory until somebody changed a setting, and renaming this one would stop it being required at all. So the step goes inside the job that already blocks merges, and the job name is left alone.

The unreachable URL is deliberate rather than lazy: if a future `drizzle-kit` starts connecting during `generate`, this fails loudly instead of quietly reading a database.

## What a green check does NOT mean

The part worth reading twice, and it is written into `CONTRIBUTING.md` and the workflow rather than left to be inferred.

The generator only sees what it can express. Measured against this database:

```
triggers:                 9
trigger functions:        5
deferrable constraints:   2

snapshots containing any of them:   none, ever
```

```
lifecycle_event_authority_insert      product_provenance_on_insert
lifecycle_event_no_delete             product_provenance_on_status_change
lifecycle_event_no_truncate           product_status_guard_insert
lifecycle_event_no_update             product_status_guard_update
lifecycle_event_recorded_at_is_ours
```

Every guarantee this project has argued about for a week lives in that list. A green drift check means **"the part of the schema drizzle-kit can represent is not stale"** — nothing more. The triggers are covered by the integration suite against a real Postgres, which is a separate job for exactly this reason.

Reading the check as "Postgres matches the schema" would be a worse state than having no check, because it would be a reassurance nobody verified.

## Tests

No new tests. The change is a migration that does nothing and a CI step, and the CI step's verification is the two-direction table above rather than a unit test — a test asserting that `drizzle-kit` emits nothing would be testing `drizzle-kit`.

The existing 474 pass against a from-zero database, which is the real assertion that `0017` is correct.

## Security impact

None directly. Indirectly: a migration pipeline people distrust is one they work around, and the guarantees in this database are all migrations.

## Breaking changes

None. `0017` is empty by design and runs as a no-op on fresh and existing databases alike.

---

- [ ] CI is green
- [x] Documentation updated: `CONTRIBUTING.md`, the workflow comment, and `0017` itself


clevervi:

Labelled per `CONTRIBUTING.md`. Two accounts authored this branch and that is authorship, not review: `e273f74` and `eea826c` are clevervi's, `f9e05d6` is Raishark's. Neither approved the other's work.

### The check proved itself in CI, which is the only proof that counts here

Step 9 of the required `Lint, typecheck and build` job:

```
9. Migrations describe the schema drizzle-kit can see -> success
   No schema changes, nothing to migrate 😴
```

With **no `DATABASE_URL` anywhere in that job**. That is simultaneously the evidence for three separate claims: the reconciliation in `0017` holds, `generate` genuinely needs no database, and the step sits inside a check that blocks merges rather than beside one that does not.

### SonarCloud was right, and I asked it this time

It rated the branch **E on security**:

```
[BLOCKER] secrets:S6698  .github/workflows/ci.yml:69
Make sure this PostgreSQL password gets changed and removed from the code.
```

My first instinct was that this is a false positive — the URL pointed at `127.0.0.1:1`, deliberately unroutable, and nothing could ever authenticate with it.

**That instinct was wrong, and the rule is better than my reasoning about it.** A scanner cannot know a credential is fake, the next person copying the line cannot either, and "it is fine because it does not work" is an argument that stops being true the moment someone makes it work. The right answer was never to suppress the rule; it was to stop needing the credential — which meant fixing `drizzle.config.ts`, which had been demanding one for a command that never connects. The workflow was the symptom.

Worth recording alongside the previous pull request, where I *guessed* at a Sonar failure and blamed the wrong file. This time I queried the API for the actual rule and line before touching anything. Two failures, two different lessons, same root habit.

**1. `process.argv[2] !== "generate"` is a heuristic, and heuristics rot.**

It reads the subcommand position, which is more precise than searching all of `argv` — verified that a `migrate` run from a path containing the word "generate" still requires a credential, where `includes` would have waived it. The default is to require, so an unknown command errs toward asking.

But it is still coupled to how `drizzle-kit` shapes its arguments. If a future version accepts global flags before the subcommand, `argv[2]` stops being the command and every call starts demanding a credential. That is the safe direction to fail in, which is why it is acceptable, not why it is right.

**2. The no-credential path is untestable on this machine, and CI is doing the work.**

The root `.env` supplies `DATABASE_URL` to every local run, so the branch this fix exists for cannot be exercised here. I did not move or edit `.env` to force it. The CI run above is the real test — which is fine, but it means a local `pnpm db:generate` will never tell you whether this guard is intact.

**3. The `test` job still hardcodes `trustpass_ci`, and I left it.**

Same rule, same file, sixteen lines down. Sonar ignored it because it is not new code.

Deliberate: that container is ephemeral, exists only inside that job, and the identical value already appears in the service definition directly above it. Moving it to a repository secret would protect nothing and would *suggest* it protected something, which is worse than leaving it visible. Said out loud here so it reads as a decision rather than an oversight — and if someone disagrees, it is a two-line change.

### The limit, restated because it is the part that will be forgotten

A green drift check means **the part of the schema drizzle-kit can represent is not stale**. Measured: nine triggers, five trigger functions, two deferrable constraints, and no snapshot has ever contained one of them. Every guarantee argued about this week lives in that blind spot.

`CONTRIBUTING.md` and the workflow comment both say so. If this check ever gets described as "the schema is verified", that is the sentence to point at.

Issue #89, opened 2026-09-18
[FRAUD] TP-111: the same serial under two different issuers is a signal nothing currently weighs


clevervi:
## Why this exists

`packages/db/src/schema/product.ts:194` already routes a decision here:

> Scoped to one issuer on purpose. Two issuers holding units with colliding serials is legitimate -- serials are only unique within a manufacturer's own numbering. The same serial across different issuers is a fraud signal to weigh, not a constraint to enforce (TP-111).

`TP-111` appears in that comment and **nowhere else**. `docs/ROADMAP.md` assigns `TP-010` through `TP-141` and contains no `TP-111`; the fraud work exists only as `TP-EPIC-11`, unnumbered.

So the code defers a real decision to an identifier that owns nothing. [#49](https://github.com/clevervi/trustpass/issues/49) opens with the reason that matters: *"An acknowledged problem with no owner is a problem that disappears."*

Found while writing ADR 0010, which routes behaviour here too — a correspondence proposed between two **issuer** records asserts that two manufacturers made the same unit, which is a conflict rather than a join.

## The situation

`product_live_issuer_serial_idx` is `(issuer_id, lower(serial))` where `status <> 'retired'`. Deliberately per issuer, and correctly so: serials are unique only within a maker's numbering, so two makers using `SN-00001` is ordinary and refusing the second would break honest registrations.

That leaves the collision visible and unexamined:

```
issuer A  ──▶  serial SN-00001   (legitimate)
issuer B  ──▶  serial SN-00001   (legitimate — or a counterfeit claiming A's numbering)
```

Nothing tells those apart today, and nothing even records that the pair occurred.

## What this is not

**Not a constraint.** Enforcing global serial uniqueness is the wrong fix and would break real registrations. That was decided when the index was scoped and is not being reopened.

**Not a verdict.** Per ADR 0003 the system surfaces what it knows; a collision is evidence to weigh, not a finding to publish. A passport must never say "counterfeit" because two issuers used the same string.

## Acceptance criteria

- [ ] A collision across issuers is detected and recorded when it occurs, not discovered by a query somebody thought to run.
- [ ] Recorded as an observation, with both records named — never as a status change on either, since neither product did anything wrong.
- [ ] Decide what, if anything, the public passport shows. The default is nothing: the honest options are "say nothing" and "say two records share this serial", and "flag it" is not among them.
- [ ] `docs/ROADMAP.md` assigns `TP-111` so the reference in `product.ts` resolves to something.
[...]

PR #90, opened 2026-09-18
docs: two records for one object, joined by evidence

### LINKED ISSUE #49: [IDENTITY] Reconcile a holder enrolment with a later manufacturer registration

clevervi:
## What changed

ADR 0010 — how two TrustPass IDs for one physical object are reconciled. No code. Registered in the ADR index and `CHANGELOG.md`.

## Why

Closes #49

A person enrols a graphics card. Later its manufacturer registers the same serial. `product_live_issuer_serial_idx` is scoped per issuer and `product_live_holder_serial_idx` to records with none, so **both rows are legal and neither index fires**.

#49 already ruled out all three obvious answers, and this ADR is built on that reasoning rather than repeating it: merging kills an identifier already printed on an object, separating gives a buyer a partial passport that looks complete, and "the manufacturer claims it" needs an answer to how a manufacturer proves the unit is theirs.

## The reading this ADR exists to refuse

ADR 0009 decided that a capacity is granted, scoped and evidenced. It is tempting — and I said something close to it when proposing this work — to conclude that #49 is therefore settled: the manufacturer's claim is backed by a granted capacity, the holder's is self-asserted, so the manufacturer wins.

**Raised in review, and the objection is right.** Authority over a *class* of claims is not evidence about a *particular object*:

```
"this organisation really is the manufacturer"        <- ADR 0009 answers this
"this serial really belongs to this physical unit"    <- nothing answers this
```

A verified manufacturer typing a serial into a form has proved who it is and has proved nothing about the object in front of anyone. The asymmetry between the two assertions is real and useful, but it is a difference in **standing**, not a substitute for evidence.

So §4 enumerates evidence and states that standing is never one of the rows, and §7 refuses self-verification — otherwise "manufacturer wins" returns wearing a process, and anyone obtaining an issuer capacity can absorb records they have never seen.

Worth noting that #49 had already named this: *"it needs an answer to how does a manufacturer prove this is the unit they made — which is the same serial-matching problem, from the other side."* The issue was ahead of my summary of it.

## The dimension neither previous ADR had

```
holder:        "I have had this object since March 2026"
manufacturer:  "we made the unit carrying serial X in 2024"
```

Both true, covering different spans, contradicting nothing. A model recording only who said what and with what authority cannot express the thing that makes these two records compatible.

So every assertion carries the period it covers, and a verified correspondence **bounds** the unknown period rather than erasing it: the object existed from the manufacturer's date, and TrustPass knew nothing about it until the holder's. That smaller, honestly described unknown is the actual benefit of reconciling — and the reason this is worth building rather than routing around.

## The questions it answers unambiguously

Set as the bar in review. Each is decided in the document:

| Question | Answer |
| --- | --- |
| What does it mean that two IDs are candidates for one object? | A correspondence exists between them, in state `proposed` |
| What evidence can join them? | An enumerated set. **Serial match alone is not enough** |
| Who may propose? | An issuer over its own products, a holder over their enrolment, or the system on a collision |
| Who may approve? | Never the proposer alone; the counterparty's side confirms |
| What if rejected? | Both records continue unchanged, and the rejection is **kept** — repeated rejections are a pattern |
| Do both IDs still resolve? | Yes, forever, and each names the other once verified |
| Is the unknown period preserved? | Yes. Bounded at one end, never erased |
| Can it be reversed? | Yes, to `withdrawn`, as an event. Nothing is deleted |
| Two manufacturers, one serial? | **Not a correspondence.** A conflict, routed to #89 |

## Two references in my own draft pointed at nothing

Caught by checking rather than by review, which is the only reason they are not in the merged document.

**`TP-130`** was cited twice for secure tags. `docs/ROADMAP.md` assigns `TP-010` through `TP-141` and contains no `TP-130` — NFC exists only as `TP-EPIC-13`, unscheduled. Corrected to say so.

**`TP-111`** was cited as where a two-issuer serial collision becomes a fraud signal. It appears in `packages/db/src/schema/product.ts:194` and **nowhere else** — not in the roadmap, not as an issue. So the code was already deferring a decision to an identifier that owned nothing, which is precisely the failure #49's own opening paragraph names:

> An acknowledged problem with no owner is a problem that disappears.

Opened as **#89** rather than corrected into vagueness, so both the existing code comment and this ADR now point at something real.

Every remaining `TP-` id in the ADR was checked against the roadmap.

## Tests

None. No code changed.

The honest limit, same as ADR 0009: an ADR is not testable and this one is load-bearing. It describes a model that does not exist yet, and the first implementation will find things the document got wrong. What it buys is that those will be arguments about a written decision rather than about whatever the first pull request happened to do.

## Security impact

Nothing is enforced by this PR. It names a new thing worth attacking — a verified correspondence moves trust between records, so proposing them in volume, or obtaining an issuer capacity in order to propose one, becomes worthwhile. That is why §7 refuses self-verification, and it gives `TP-100`…`TP-105` another object to protect.

**None of this can be built before `TP-141`.** "The holder confirms" and "the issuer confirms" are both meaningless while `actor_kind` is self-declared, and the ADR says so rather than leaving it to be discovered.

## Breaking changes

None now. States a direction that is expensive later: correspondences are events, and events are permanent, so a wrong shape here is migrated rather than edited.

---

- [ ] CI is green
- [x] `pnpm lint` passes locally; typecheck, test and build unaffected — no code changed


clevervi:

Labelled per `CONTRIBUTING.md`. One person wrote this and one person is reading it, and on an ADR that is worth even less than usual: there is no suite to disagree with me and no mutation that can kill a paragraph.

So here are the places I think this is weakest.

### 1. §7 refuses self-verification, and that may make the common case impossible

The rule is that a correspondence reaches `verified` only when the counterparty's side confirms — the holder confirms a manufacturer's proposal, and an issuer confirms a holder's.

I believe it is right, because the alternative hands anyone with an issuer capacity the ability to absorb records they have never seen.

**But consider the ordinary case it blocks.** A manufacturer proposes a correspondence against a card someone enrolled two years ago. That person has sold the card, changed their email, or simply does not care. Nobody confirms. The correspondence sits in `proposed` forever, the buyer scanning the code still sees two unconnected passports, and the feature helped nobody.

The ADR has no answer for that, and I did not write one in because every answer I could think of was worse: a timeout that auto-verifies is self-verification with a delay; an administrator who verifies on the holder's behalf is a person with the power to join any two records. **A rule that is correct and unusable is still a design failure**, and this is the part most likely to change on first contact with a real case.

### 2. "A correspondence is a record, not an edit" is clean and possibly too clean

Every read of a passport now has to resolve correspondences, render whichever states apply, and say what each side covers — without any of `proposed`, `verified`, `rejected` or `withdrawn` sounding like a verdict. That is ADR 0003's whole problem arriving somewhere new, and `claim-wording.ts` is already the most carefully argued file in the repository.

The ADR hands that cost to the implementation and does not price it. I think the alternative is worse, but "we chose the honest model" is easier to write than to render.

### 3. The evidence table has four rows and only one strong one, which does not exist

`Serial match` is admitted to be worth very little. `Holder confirmation` is weak alone. `Manufacturer production record` is an opaque reference and ADR 0008 already says a reference is not evidence that was checked. The only strong row is a secure tag, which is `TP-EPIC-13` and unscheduled.

So this ADR describes a mechanism whose evidence is, today, **three weak things**. That is honest and it is also a real limit: a correspondence verified on serial-match-plus-holder-confirmation is a better guess than either record alone, not a proof. The document says as much. Whether that is worth the machinery is a judgement somebody should be able to disagree with, so it is stated plainly rather than buried under the process.

### 4. Two references in my own draft pointed at nothing

`TP-130` is not in the roadmap at all. `TP-111` was in a code comment and nowhere else. I caught both by checking every `TP-` id against `docs/ROADMAP.md` before pushing — not by reading, which had already passed over them twice.

Recording it because the lesson is the same one this session keeps producing: the claims that survive review are the ones nobody thought to verify. `TP-111` now has an owner (#89) instead of being quietly softened into prose.

### What I am confident about

**Standing is never evidence.** That one is not a preference. Recording "a verified manufacturer said so" as though it were evidence about the object is exactly the collapse ADR 0003 exists to prevent, reached through a new table instead of through a badge.

**Both identifiers resolve forever.** A code printed on a physical object that stops working is the failure the whole identity design was built around, and no reconciliation is worth reintroducing it.

PR #103, opened 2026-09-18
docs: authority is pinned to the moment it was used

### LINKED ISSUE #102: [SECURITY] TP-141: ADR — who may speak, and why that authority was still valid when they spoke
## Objective

The logical model for identity, credentials and capacity grants. **No tables.** The schema is a separate piece of work and writing it first is how the model gets decided by whatever the first migration happened to do.

## Where this sits

ADR 0009 decided that a capacity is granted, scoped, evidenced and revocable rather than a role on a user. ADR 0010 decided how two records for one object are reconciled, and that an authenticated actor proves **who** and never **which object**.

Neither is implemented. `lifecycle_event.actor_kind` is still self-declared: whoever reaches a write path names the capacity they are acting in and the database believes them, which is why the API must not be exposed publicly.

So the question this ADR closes is not "how do we log people in". It is:

> **Who are you, what may you assert, and why was that authority still valid when you asserted it?**

The third clause is the hard one, and it is the one a users-and-roles table answers worst.

## The invariant that makes this urgent

Events are permanent (ADR 0008). Every event written after `TP-141` points at whatever identity model exists that day, **forever**. A model that cannot express "this authority's warrant ended in March" will have recorded events that can never be re-read honestly — not lost, never captured.

ADR 0009 already says revoking a capacity does not invalidate what was recorded under it. That is currently a sentence. This ADR has to make it a **structure**, so the claim is demonstrable rather than asserted.

## What the ADR must decide

Each with the alternative rejected and what it would have cost.

- **Four things, kept apart.** Actor, Organization, Credential, Grant. Specifically **not** `organization.type = manufacturer`, which is the static-role problem returning under a new name.
- **Membership** — an actor acting on behalf of an organization, and whether that is time-bounded too.
- **An event pins the grant it acted under**, not just the capacity. `actor_kind = 'authority'` cannot answer who, or under what warrant.
- **Historical attribution is immutable.** An event records the actor and the grant that existed when it was accepted. A later revocation must not be able to rewrite it — and the ADR must say what makes that true structurally, not merely intended.
- **A credential proves the actor and grants nothing.** Authentication resolves a credential to an actor; authorisation is a lookup of that actor's grants valid **at the time of the request**.
- **Expiry and revocation** — no new claims under a lapsed grant, no invalidation of old ones. Two halves of one rule.
- **Credential compromise and rotation**, including the uncomfortable half: what happens to actions taken with a credential before anyone knew it was compromised.
- **Scope is mandatory.** Per ADR 0009; restated because it is the rule most likely to be skipped when the first grant is written by hand.

## The limit this repository has to state about itself

"At least two distinct actors" (ADR 0010 §7) assumes two actors are two people.

**This repository is run by one person with two GitHub accounts**, and that is disclosed in `CONTRIBUTING.md`. No identity model proposed here can detect it. The ADR must say so plainly rather than let the separation-of-duties rule read as stronger than it is — the same honesty the merge bar already applies to self-review.

## Acceptance criteria

- [ ] `docs/adr/0011-*.md`, decided rather than surveyed, in the repository's ADR format.
- [ ] Every point above answered, each with the rejected alternative and its cost.
- [ ] States what makes historical attribution immutable **structurally**, since asserting it is what ADR 0009 already did.
- [ ] Says explicitly what it does **not** decide, so the implementation has a scope rather than an invitation: no credential mechanism, no session handling, no tables.
- [ ] Names the two-accounts limit.
- [ ] Registered in `docs/adr/README.md` and `CHANGELOG.md`.

## Risk

`critical`. Events will reference this model permanently. A wrong shape is not a refactor, it is a migration of history.

clevervi:
## What changed

ADR 0011 — the logical model for `TP-141`. **No tables.** Registered in the index and `CHANGELOG.md`.

## Why

Closes #102

ADR 0009 decided that a capacity is granted, scoped, evidenced and revocable. ADR 0010 decided that an authenticated actor proves **who is speaking** and never **which object**. Neither is implemented, and `actor_kind` is still self-declared.

So the question is not how to log people in:

> **Who are you, what may you assert, and why was that authority still valid when you asserted it?**

The third clause is where this gets hard, and it is the one a users-and-roles table answers worst.

## The decision worth arguing about

ADR 0009 says revoking a capacity does not invalidate what was recorded under it. **That was a promise in a document.**

A model where revocation *could* rewrite history but everyone agrees not to is not a model. It is an etiquette.

So **grants are append-only**, exactly like `lifecycle_event` and for the same reason. Never updated: not to revoke, not to extend, not to rescope. Every change is a new record. *"What authority did this actor hold on 2 August"* becomes a **query**, not a belief, and reconstructing it does not depend on anyone having refrained from an `UPDATE`.

### The alternative that will be proposed first

A `revoked_at` column on the grant. One column, and it is the whole problem:

```
event  ──points at──▶  grant
                        └─ revoked_at set later

The event never changed. The grant changed underneath it.
History rewritten without a single event being touched.
```

Also rejected: copying the grant terms onto the event. It survives the edit and creates two copies that can disagree, with nothing to test them against — unlike `product-status`, where both definitions are compared against each other by an integration suite.

## Four things, kept apart

| | Answers | Lifetime |
| --- | --- | --- |
| **Actor** | the thing that acts | permanent |
| **Organization** | which legal party it acts for | permanent |
| **Credential** | how it proves it is that actor | expires, rotates, is revoked |
| **Grant** | what it may assert, over what, from when, until when | revoked, never edited |

Specifically **not** `organization.type = manufacturer`. That is the static-role problem returning under a new name: one type and two jobs the day a maker also runs a service centre, an array the day it gains a third, and the scoping gone.

And an event pins **the grant**, not the capacity. `actor_kind = 'authority'` cannot say which authority — the schema currently *requires* it to stay silent, since `lifecycle_event_issuer_matches_actor` forces `issuer_id IS NULL` unless the actor is an issuer.

## The uncomfortable half, decided rather than left

A compromised credential is revoked and the actor survives. But actions taken **before anyone knew** it was compromised **stand as recorded**.

The intuitive answer — void them retroactively — deletes legitimate records, because most actions in that window were the real actor's. It also hands anyone who can claim a compromise a way to erase their own history.

## The limit this repository has to state about itself

ADR 0010 §7 requires at least two distinct actors across proposer, evidence provider and verifier. **That rule assumes two actors are two people.**

This repository is run by one person with two GitHub accounts, disclosed in `CONTRIBUTING.md` and visible in the history. **No model in this document can detect that.** Two actors, two credentials, two memberships, one human deciding both.

So separation of duties here is a **procedural** control, not a technical one — the same honesty the merge bar already applies to self-review. Letting the rule read as enforced would be the first false claim in the model.

## What it deliberately does not decide

So the implementation has a scope rather than an invitation: the credential mechanism (password, API key, mTLS, OIDC, passkeys), session handling, token lifetime, **the schema**, how an organization gets verified in the first place, whether grants are delegable, and the public wording.

## Tests

None. No code changed.

The honest limit, third ADR running: this is not testable and it is load-bearing. Every event written after `TP-141` will point at this model permanently, because events are permanent. What it buys is that the first implementation's mistakes will be arguments about a written decision rather than about whatever the first pull request happened to do.

## Security impact

Nothing is enforced. It names what becomes worth stealing: not a credential — credentials rotate — but the **grant** behind it, which turns `TP-100` … `TP-105` from generic hardening into the defence of one specific record.

## Breaking changes

None now. Expensive later: events will reference grants, and a grant referenced by an append-only history cannot be deleted, only ended.

---

- [ ] CI is green
- [x] `pnpm lint` passes locally
- [x] Every `TP-` identifier checked against the roadmap per the rule in `CONTRIBUTING.md`


clevervi:

Labelled per `CONTRIBUTING.md`. `77bb680` is clevervi's, `ab63173` is Raishark's; that is shared authorship, not review.

And this ADR is the one where that distinction stops being a formality, because §"the limit this repository has to state about itself" is about exactly this arrangement. A document arguing that separation of duties here is procedural rather than technical, reviewed by the person who wrote it, using the second account.

**1. Append-only grants put a history query in the hot path of every write.**

"The grants this actor currently holds" stops being `WHERE revoked_at IS NULL` and becomes a reduction over every grant record for that actor. Every write path does it. Every one.

I think it is right — §4's argument is that a mutable grant makes the central guarantee unprovable, and an unprovable guarantee is not one. But **"correct and slow" has a way of becoming "correct, slow, and quietly cached into a mutable projection"**, which is the same defect with a performance justification. If this is undone later, that is how.

The ADR does not price it. It could not, honestly, without a schema — but it should not read as though the cost were hypothetical.

**2. Four separate things is more model than today's problem needs.**

Nothing in this system has more than one organization, one capacity per actor, or a credential of any kind. The four-part split is justified against problems that have not arrived: an employee leaving, a maker that also repairs, a rotated key.

That is the same reasoning I flagged as weakest in ADR 0009 §2 and it has the same answer — events are permanent, so the model must be able to express what an event will need to say, and a fact nobody captured cannot be migrated back. I still think it holds. It is still a bet on predicted needs, made by someone who has not met the real ones.

**3. "Actions before a known compromise stand as recorded" is the decision I am least sure of.**

The argument is sound: most actions in that window were the real actor's, and voiding them hands anyone who can claim a compromise a way to erase their own history.

But the case it loses is real and unpleasant — a stolen credential used to suspend somebody's product, and the record stands with a qualifier while the owner is told the reason it happened was a theft of somebody else's key. The ADR makes the right trade and it **is** a trade, not a free choice, and nothing in the document currently says what the qualifier should look like to the person on the receiving end.

### What I am confident about

**The `revoked_at` column rejection.** Not a preference. It is one line, it looks equivalent to the append-only form, and it makes the guarantee unprovable — the event never changes while the grant changes underneath it, so history is rewritten without an event being touched. That is the rejection most likely to be quietly reversed during implementation, which is why the changelog entry leads with it rather than with the four-part model.

**Stating the two-accounts limit.** ADR 0010 asks for two distinct actors; this repository cannot supply them and no model here can tell. Writing that down costs nothing and not writing it would have made the separation-of-duties rule the first false claim in a document about proving who may speak.

PR #107, opened 2026-09-18
feat: authority becomes a record with a lifetime

### LINKED ISSUE #106: [SECURITY] TP-152: the relational and temporal shape of authority, decided before the SQL
## Scope

The schema for ADR 0011. **Decided here before any `CREATE TABLE`**, because the table names matter far less than the relational and temporal invariants, and the invariants are what a migration freezes.

ADR 0011 rejects a `revoked_at` column on the grant and says every change is a new record. It deliberately names **no structure** for that — this issue picks one.

## The question the shape has to answer

> What authority did this actor hold at 14:02 on 2 August?

Reconstructible, not inferred, and not dependent on anyone having refrained from an `UPDATE`.

## Decided: revocation is its own fact

```
grant                       grant_revocation
├── id                      ├── grant_id      (unique — a grant is revoked once)
├── actor_id                ├── revoked_at
├── organization_id         ├── revoked_by    (actor)
├── capacity                ├── reason
├── scope                   └── evidence_reference
├── effective_from
├── expires_at   (nullable = no fixed end)
├── granted_by   (actor)
└── evidence_reference
```

Neither table is ever updated. Both are append-only, enforced by trigger as `lifecycle_event` already is, because a rule living only in application code is bypassed by the first path that writes without going through it.

**Extending or rescoping is not a mutation and not a revocation — it is a new grant.** ADR 0011 §4 says so; the shape has to make it the only option, which it does by having nowhere to put an edit.

*Rejected: a generic `grant_event` table.* More general and it buys nothing here. A grant has exactly one revocation and no other transitions, so a table modelling arbitrary ones would be an empty abstraction with a `type` column nobody switches on.

## Decided: the temporal query, and the clause that is easy to get wrong

```sql
SELECT g.*
FROM grant g
LEFT JOIN grant_revocation r ON r.grant_id = g.id
WHERE g.actor_id = $actor
  AND g.effective_from <= $t
  AND (g.expires_at  IS NULL OR g.expires_at  >  $t)
  AND (r.revoked_at  IS NULL OR r.revoked_at  >  $t)
```

The last line is the whole point and the one a reasonable implementation gets wrong:

> **`r.revoked_at > $t`, not `r.grant_id IS NULL`.**

A grant revoked in September **was valid in August**. Filtering out every revoked grant answers "what may this actor do now" and silently destroys the answer to "what could they do then" — which is the question ADR 0011 exists to keep answerable. A test must fail if that clause degrades.

## Decided: `grant_id` on `lifecycle_event`, and what NULL means

Nullable. There are 8,000+ existing events and no grants, because authority did not exist when they were recorded.

> **`NULL` means "recorded before TrustPass could prove authority". It does not mean "recorded without authority".**

Written into the column comment and pinned by a regression test, because the misreading is one `WHERE grant_id IS NULL` away and it turns a limitation of the system into an accusation about a record. That is ADR 0003's failure reached through a nullable column.

*Rejected: a sentinel "legacy" grant.* It would have 8,000 events pointing at a grant nobody issued, to nobody, for nothing — inventing a continuity that did not exist, which is the same objection ADR 0010 makes to merging two records.

## Decided: the deployment order, because the constraint cannot come first

Raised in review. A trigger requiring `grant_id IS NOT NULL` would break every existing write path the moment it lands, since none of them resolve a grant yet.

```
this issue    schema + temporal query + tests        grant_id nullable, no trigger
next          write paths resolve and persist grants
then          trigger refuses NULL on new events     fail closed, forward only
last          actor_kind demoted to derived
```

The trigger is deliberately **not** in this phase. A constraint that arrives before the code that satisfies it is an outage in a project with deployments and a broken suite in one without.

## Acceptance criteria

- [ ] `actor`, `organization`, `membership`, `credential`, `grant`, `grant_revocation`.
- [ ] `grant` and `grant_revocation` append-only by trigger, with their own SQLSTATE, distinct from `TP002`.
- [ ] `lifecycle_event.grant_id`, nullable, FK restrict, with the NULL semantics in the column comment.
- [ ] The temporal lookup as a tested function.
- [ ] Every guard mutation-checked: break it, watch a named test go red, restore it.

### Cases that must be tested from the first commit

Raised in review, and each proves a boundary the model would otherwise only claim:

- [ ] **A valid credential with an expired grant authorises nothing.** Authentication is not authorisation, and this is the cheapest possible demonstration.
- [ ] **A valid credential with a revoked grant authorises nothing.**
- [ ] **A grant revoked in September still answers "valid" for a query at August.** The clause above.
- [ ] **A rotated credential does not change any historical event.** The event points at the grant, never at the credential in force today.
- [ ] **`grant_id IS NULL` on a legacy event is not an absence of authority.** The regression test for the misreading.
- [ ] A grant with no scope is refused.

## Explicitly not in this phase

Authentication mechanism — passkeys, OIDC, magic links, API credentials, signatures. ADR 0011 §5 constrains what a credential is *for*, not what it is made of. **The mechanism is interchangeable; the historical identity is not.**

## Risk

`critical`. Events will reference grants permanently. A wrong shape is migrated, not refactored.

clevervi:
## What changed

ADR 0011's model, as tables. `actor`, `organization`, `membership`, `credential`, `capacity_grant`, `capacity_grant_revocation`, and `grant_id` on `lifecycle_event`.

Nothing authenticates yet. This is the thing authentication will resolve *to*.

## Why

Closes #106

`actor_kind` is self-declared: whoever reaches a write path names the capacity they are acting in and the database believes them. These tables are what it will be checked against.

## The shape that matters: `capacity_grant` has no `revoked_at`

**The absence is the design, not an omission.** ADR 0011 §4 rejects the column explicitly:

```
event  ──points at──▶  grant
                        └─ revoked_at set later

The event never changed. The grant changed underneath it.
History rewritten without a single event being touched.
```

So a revocation is its own fact, one per grant by unique index, and both tables are append-only by trigger under a new `TP005` — distinct from `TP002`, because a caller needs to tell *"the past cannot be edited"* from *"the authority for the past cannot be edited"*.

Extending or rescoping is not a mutation and not a revocation: it is a **new grant**, and the shape makes that the only option by having nowhere to put an edit.

## The clause an implementation gets wrong

```sql
AND (r.revoked_at IS NULL OR r.revoked_at > $t)   -- correct
AND  r.revoked_at IS NULL                          -- the plausible mistake
```

A grant revoked in September **was valid in August**. The second form answers *"what may this actor do now"* and silently destroys the answer to *"what could they do then"* — the question ADR 0011 exists to keep answerable. The two look identical from the call site.

So `grantsHeldAt(db, actorId, at)` takes the instant as an argument rather than defaulting to `now()`. An authorisation check asks about this moment; a passport rendering a two-year-old event asks about that one.

## Measured before naming anything

`grant` is a **reserved word**:

```
CREATE TABLE grant (id int);
ERROR:  syntax error at or near "grant"

SELECT word, catcode FROM pg_get_keywords() WHERE word = 'grant';
 grant | R
```

Quoting works, and would put `"grant"` inside every trigger body in the migration, where one missing pair of quotes is a syntax error buried in PL/pgSQL. Hence `capacity_grant`.

## Two decisions that come straight from earlier ADRs

**`holder` cannot be granted**, by check constraint. ADR 0009 §8: a holder's capacity is granted by nobody, because no registry of people who own things exists and inventing a grant would fabricate an authority. A grantable `holder` would make that honest carve-out silently untrue.

**No grant is unscoped.** `own_organization` requires an organization, `country` requires a country, both by check. ADR 0011 §8: a capacity with no scope is not a permission, it is a master key.

## `grant_id` is nullable, and what NULL means

> **"Recorded before TrustPass could prove authority."** Not "recorded without authority."

8,000+ events predate the column and no grant existed to point them at. A "legacy" grant would have them referencing something nobody issued, to nobody, for nothing — the same objection ADR 0010 makes to merging two records to tidy them up.

The misreading is one `WHERE grant_id IS NULL` away from turning a limit of the system into an accusation about a record, so the semantics is in the **column comment** — where somebody reading the database instead of the repository will meet it — and asserted by a test rather than trusted to survive a future migration.

**No trigger requiring it yet, deliberately.** No write path resolves a grant, so a constraint arriving before the code that satisfies it is a broken suite here and an outage anywhere with deployments. The order is: schema, then write paths, then the trigger.

## An import cycle, resolved by moving one declaration

`capacity_grant` needs the capacity enum; `lifecycle_event` needs a foreign key to the grant. The enum moved to `actor-capacity.ts`, which breaks the cycle without renaming the Postgres type — a rename buys nothing and the old name is already referenced by a trigger function and by the `RECORDING_AUTHORITY` suite.

## Tests

18 added, **492 total** (121 web, 311 db, 60 api). All four gates green, plus the migration drift check from #86.

### Mutation table

| Mutation | Tests that went red | Result |
| --- | --- | --- |
| `revokedAt > at` → `revokedAt IS NULL` | `answers valid for August when the revocation lands in September` | `1 failed \| 310 passed` |
| Drop the append-only triggers on `capacity_grant` | 4 authority-rewrite tests, each naming `TP005` | `4 failed \| 307 passed` |
| `membership_only_ends()` returns `NEW` unconditionally | `refuses moving an actor to a different organization`, `refuses deleting one` | `2 failed \| 309 passed` |

The first is the one worth noting: it killed **exactly one** test, and that test is the one named for the clause. Nothing else moved — a guard whose removal breaks unrelated tests is either too broad or is being measured by accident.

### The four cases that prove authentication is not authorisation

- A valid credential with an **expired** grant authorises nothing
- A valid credential with a **revoked** grant authorises nothing
- An actor with a credential and **no grant** may do nothing at all
- A **rotated** credential changes nothing about what was held

The last matters most. If rotation could move authority, every historical attribution would be re-decided by whatever key the actor happens to hold today.

Plus the boundary at the instant of revocation, because an off-by-one there decides whether an action taken in the same second as a revocation was authorised.

## Security impact

This is the foundation of it, and none of it is enforced yet. `actor_kind` remains self-declared until write paths resolve grants, and the API must still not be exposed publicly.

What the tables make possible: an event that can prove **who** acted, **why** they were allowed to, and **whether that was still valid at the time** — none of which is answerable today.

## Breaking changes

Schema: additive, forward-only. Six new tables, one nullable column, one new SQLSTATE. `0018_authority_is_a_grant`.

---

- [ ] CI is green
- [x] `pnpm db:generate` produces nothing — the snapshot is in step
- [x] Documentation updated: the migration's reasoning, every table's doc comment, and the column comment


clevervi:

Labelled per `CONTRIBUTING.md`. `afb0dc2` is clevervi's, `eff8ea6` is Raishark's; authorship, not review.

And this is the pull request where that caveat has teeth, because ADR 0011 says so in writing: separation of duties here is procedural, not technical, and no model in the diff can detect that both accounts are one person.

### Four things worth disagreeing with

**1. `grant_history` is exported and nothing uses it.**

It exists because audit will need it and because writing it alongside `grantsHeldAt` made the difference between the two obvious — one is for authorisation and takes an instant, the other is for reading and takes none. That is a justification for a comment, not for exported code. **It is speculative surface, added on a hunch about a caller that does not exist**, and if anything here should have waited, it is this.

**2. The `own_organization` scope is declared and nothing evaluates it.**

`scope_kind` is stored and checked for coherence; no code anywhere resolves *"is this product within this grant's scope"*. So a grant currently records an intention the system cannot act on.

That is correct for this phase — the resolution belongs with the write paths — but it means the strongest claim in the ADR, that no capacity is global, is today enforced only in the sense that the column cannot be null. **Scope is real in the schema and not yet real in behaviour**, and a reader skimming the tests could easily conclude otherwise.

**3. Six tables arrive at once, for one problem that does not exist yet.**

Nothing authenticates. `credential` holds no secret and proves nothing. `membership` has no consumer. The justification is ADR 0011's — events are permanent, so the model must be able to express what an event will need to say — and it is the same bet I flagged in ADR 0009 §2 and again in ADR 0011.

Three flags is a pattern worth naming rather than repeating. **Every one of these documents has justified present complexity with future permanence.** That argument is correct here and it is also the argument that would justify anything, and I do not have a rule that separates the two beyond judgement.

**4. `capacity_grant_revocation` will need more than a `varchar` reason.**

`lifecycle_event.reason` is a closed enum, for ADR 0008's argument: free text cannot be queried and invites a sentence where a fact belongs. This table takes free text for the same kind of field, on the grounds that nobody knows the set yet.

That is how `reason` on events would have been justified too, before somebody wrote out the list. **The consistency argument points the other way from the pragmatic one**, and I picked the pragmatic one without a strong case.

### What I am confident about

**`revokedAt > at`.** It killed exactly one test under mutation, and that test is the one named for the clause. The failure it prevents is silent and total: every historical attribution re-decided by whoever holds authority today.

**No `revoked_at` column on the grant.** This is the decision an implementer will try to undo, because a column is one line and looks equivalent. The migration says why, the ADR says why, and this comment says it a third time on purpose.

**`grant_id` nullable with the meaning in the column comment.** The test asserting that comment looks like testing documentation. It is testing the only place the semantics exists for somebody reading the database rather than the repository, and the misreading it guards against turns a limitation of the system into an accusation about a record.

Issue #109, opened 2026-09-18
[IDENTITY] TP-154: issuer and organization are two tables for one kind of thing


clevervi:
## What happened

#107 introduced `organization` and did not reconcile it with `issuer`. The database now holds both, unconnected:

```
product.issuer_id        -> issuer.id
capacity_grant.organization_id -> organization.id

issuer <-> organization:  nothing
```

Two tables for "a business with a legal identity and a registration number", which is the duplication ADR 0006 refused for naming and ADR 0009 refused explicitly: *"the same real-world entity — a manufacturer that also runs its own service centre — would exist twice, with two verification states that can disagree."*

Introduced by the pull request that cited that ADR as its reason for adding the table. Named here rather than left for somebody to find.

## The immediate consequence

`capacity_grant.scope_kind = 'own_organization'` **cannot be evaluated against a product.** The grant names an organization; the product names an issuer; there is no join.

So scope is real in the schema and not yet real in behaviour, which #107's self-review already flagged — and this is why, rather than it being merely unfinished.

## Why this was not wrong to defer

Reconciling means every foreign key pointing at `issuer` moves, and `issuer` is referenced by `product` and `lifecycle_event` with `ON DELETE RESTRICT`. That is a migration with real blast radius and it does not belong inside a schema-introduction pull request.

What was wrong was not saying so.

## Options, none chosen

- **`organization` absorbs `issuer`.** ADR 0009's stated direction. Every reference moves; `issuer.verification_status` becomes a property of the organization or of a grant. Largest, and the one the ADR argues for.
- **`issuer` gains `organization_id`.** Cheapest, and it keeps two rows for one party — the defect, with a pointer.
- **Grants scope to an issuer instead of an organization.** Smallest change to make scope work, and it contradicts ADR 0009 by making the authority model depend on the table that exists rather than the concept that was decided.

The first is almost certainly right. It needs its own issue, its own migration and its own pull request, which is what this is for.

## Acceptance criteria

- [ ] One table represents a party with a legal identity.
- [ ] `own_organization` scope resolvable against a product.
- [ ] Every existing `issuer` row preserved, with its verification status, and every foreign key still resolving — migrations are forward-only and this one moves references that `RESTRICT` protects.
- [ ] ADR 0009's direction either followed or amended with a reason, not silently diverged from.

## Risk

`high`. Crosses module boundaries and touches product truth. Nothing is wrong at runtime today; what is wrong is that a decision has two representations and one of them will win by accident.

PR #113, opened 2026-09-18
feat: one party, one record — phase one, additive

clevervi:
## What changed

Every issuer becomes an organization; `product` and `lifecycle_event` gain a reference to it **beside** `issuer_id`, not instead of it. Two migrations, and **nothing is dropped**.

## Why

Refs #109. Implements ADR 0012.

`organization` and `issuer` were two unconnected tables for one legal party, so `own_organization` scope could not be evaluated against a product.

The destructive half — dropping `issuer_id`, then `issuer` — gets its own pull request, per `CONTRIBUTING.md`: *"a destructive change gets its own migration and its own pull request, so the diff that drops data is the whole diff."*

## Measured, before and after

```
issuers                          376
organizations              16 ->  392
products moved                  1563   unmoved 0
events moved                    1836   unmoved 0
verification carried    339 unverified, 37 verified   invented 0
```

Preconditions raise **before anything is written**, and a migration runs in one transaction — verified rather than assumed: a probe migration that created a table and then raised left **no table and no journal entry**.

### 1. The append-only guard refused the backfill, and it was right to

```
ERROR: A lifecycle event cannot be modified. Record a correcting event instead.
CONTEXT: PL/pgSQL function lifecycle_event_append_only()
```

An `UPDATE` on `lifecycle_event` **is** an edit to history, and the trigger cannot tell a harmless one from a harmful one — which is the point, because *"this edit is fine"* is what every harmful edit says.

Three ways out, and why the third:

| | |
| --- | --- |
| Widen the trigger to allow updates touching only `organization_id` | Narrow, checkable, and **permanent** — a standing exception is standing exploitable, for a migration that runs once |
| Leave 1,836 events unmoved, `organization_id` NULL | Honest, and phase two then cannot drop `issuer_id` without losing which party acted. The tension moves to a migration that also deletes |
| **Disable the trigger for one statement, inside the transaction** | DDL is transactional, so a rollback restores it. No window exists in which the guard was off and the transaction survived |

A postcondition asserts it came back, and a test asserts it **still works** — a disabled trigger still has a `pg_trigger` row.

### 2. `ADD COLUMN company_name NOT NULL` fails against rows that exist

The ordinary trap, and the reason the generated statement was not used as written. Added nullable, backfilled from `legal_name`, then constrained.

### 3. `0020` left a gap that opened in minutes

The write paths still insert `issuer_id` only. Within the time it took to write the invariant tests:

```
issuers with no organization                  21
products with an issuer and no organization   82
```

`0021` puts that in the **database** rather than in three call sites, for the reason this repository gives for every guarantee it owns: a rule in application code is bypassed by the first path that writes without going through it, and phase one exists *because* there will be more paths before phase two.

**Syncing on INSERT alone was not enough either.** Two issuers read `verified` while their organizations read `unverified` — which is precisely the defect ADR 0009 named when it refused two tables for one party: *"the same entity would exist twice, with two verification states that can disagree."* Changes now reach the organization, and a change to the **registry identity** is refused while both tables exist, because it would silently re-point the correspondence.

## What it deliberately does not do

**Invent a verification history.** `issuer.verification_status` is a mutable column and the only record of the fact, so the current value is carryable and a timeline is not. Reconstructing 2024 from a column overwritten since would be a plausible fiction — the one thing this project refuses to publish. Per ADR 0012 the events come next, and the history starts when they do.

**Change what the passport shows.** `read-passport.ts` is untouched. Identity, then verification events, then presentation — three layers, in that order.

**Collapse `company_name` and `legal_name`.** They look redundant and are not: a company sells as one name and signs as another. A migration that merged them would destroy information to tidy a schema.

## Tests

6 added, **503 total** (121 web, 322 db, 60 api). All four gates green, drift check clean.

### The invariants earned their keep on the first run

Three failed immediately and were right to — they found gap #3 above. A fourth found the verification divergence. They assert against **whatever is in the database**, not a fixture they built: a migration tested only against `db:reset` is tested against zero rows, where *"move every reference"* is trivially correct.

Each asks two questions where one would pass for the wrong reason: *"every product has an organization"* and *"it has the **right** one"* are different, and a backfill joining on the wrong column satisfies the first.

### Verified after two full suite runs, with new data written throughout

```
issuers=42  orgs=44
issuers_without_org=0  products_out_of_step=0
events_out_of_step=0   verification_disagreeing=0
```

### One of my tests was wrong

It issued an unqualified `UPDATE` to prove the append-only guard still works. On a freshly reset database that touches **zero rows** — the row trigger never fires, nothing raises, and the test fails for the one reason unrelated to the guard. It now writes its own event first.

## Security impact

The append-only guard is deliberately disabled for one statement in each migration. That is the highest-risk line in this change, and it is why the postcondition and the test exist. After both migrations the guard is enabled and refusing edits, asserted twice.

## Breaking changes

Schema: additive. Two columns, two indexes, two foreign keys, four triggers, one constraint. `0020_one_party_one_record` and `0021_the_two_identities_stay_in_step`.

The triggers in `0021` are **temporary and their end is scheduled**: when `issuer` goes, they go with it.

---

- [ ] CI is green
- [x] Applied against a database with historical data **and** from zero
- [x] `pnpm db:generate` produces nothing
- [x] Documentation updated: both migrations carry their reasoning, including the three failures


clevervi:
## SonarCloud is red, and it is reporting the thing this pull request exists to end

```
4.4% Duplication on New Code  (required ≤ 3%)

 90.9%  new_lines=33  packages/db/src/schema/organization.ts
```

Thirty of thirty-three new lines in `organization.ts` duplicate `issuer.ts`: the same `company_name`, `legal_name`, `registration_number`, `country` and `verification_status`, the same checks, the same unique index on the registry identity.

**That duplication is not accidental and it is not incidental — it is the defect.** ADR 0012 exists because two tables describe one legal party, and this pull request is the first half of removing one of them. The metric is measuring the intermediate state of the fix.

### What I am not doing about it

**Not widening `sonar-project.properties`.** The exclusion there covers migration SQL, for a reason written in the file. Adding a schema file to it would trade a working rule for a green number, which was refused on #84 and #86 and is refused here.

**Not extracting shared column definitions.** Two tables that are about to become one do not need a shared base; the extraction would be written this week and deleted in phase two, and the intermediate abstraction would make the migration harder to read while it matters most.

**Not merging past it silently.** The required checks on `develop` are `Lint, typecheck and build`, `Test against Postgres` and `Secret scan` — all green. SonarCloud is advisory here, and proceeding past a red advisory deserves a written reason rather than a shrug, which is what this comment is.

### When it goes away

Phase two drops `issuer_id` from `product` and `lifecycle_event`, then drops `issuer`. At that point `organization.ts` is the only definition of a party and the duplication is zero, because there is one table instead of two.

If this number is still above the gate after that pull request, something is wrong with the plan rather than with the metric.

Issue #122, opened 2026-09-18
[OPERATIONS] TP-164: a backup nobody has restored is a claim, not a capability


clevervi:
## Why now

`0023` dropped a table. After a migration like that, the previous identity is not recoverable from the running database — only from a backup.

There is no backup. There is no restore. `pnpm db:reset` destroys the volume and rebuilds from migrations, which recovers the **schema** and none of the data.

The rehearsal in #118 covered the migration being correct. It did not cover the case where it is correct and something else goes wrong afterwards.

## The distinction that matters

```
pg_dump ran and a file exists   →  a claim
that file restored into a database, and the queries work  →  a capability
```

Raised in review, and it is the same standard this repository already applies to tags: `CONTRIBUTING.md` says a tag is a claim that a version works and must be cut only after the artefact has been **run**, not merely built. A backup deserves the same sentence.

## Acceptance criteria

- [ ] A documented way to take a backup, and a documented way to restore one.
- [ ] A restore actually performed into a temporary database, with the passport read path and the full suite run against the result — the shape #118 used for the migration rehearsal.
- [ ] The restore verified by content, not by exit code: row counts, a sample TrustPass ID resolving with its history, verification states intact.
- [ ] Written where the other database procedures live.

## Added after review, and after #119/#127 changed what "restored" has to mean

Four things the original criteria left out. The first is the one that would have made this issue produce a false green.

**1. The roles are not in the dump.** `pg_dump` writes a database; `trustpass_owner`, `trustpass_migration` and `trustpass_runtime` are cluster-level objects, and their memberships and grants are not database rows. A restore that recovers every product and lands in a database where the runtime is an owner again has restored the data and thrown away #119 and #127 — and every test in `least-privilege.integration.test.ts` would then be describing a database that no longer exists.

- [ ] The restore procedure states how roles and their memberships come back (`pg_dumpall --roles-only`, or recreated by running `0024`), and the verification asserts the three roles exist with the right attributes.
- [ ] `least-privilege.integration.test.ts` passes against the **restored** database, connecting as `trustpass_runtime`. That is the assertion proving the privilege boundary survived, and it needs the restored database to have a provisioned runtime password.
- [ ] `assertConnectionIsUnprivileged` passes against the restored database, which is the same question asked from the application's side.

[...]

Issue #120, opened 2026-09-18
[SECURITY] TP-162: a 409 from /enrolments tells a stranger which serials are enrolled


clevervi:
## What it answers

`apps/api/src/routes/enrolments.ts`:

```
Serial ${body.serial} is already enrolled as ${result.existingTrustpassId}.
```

Post a serial, and the response says whether it is enrolled **and gives the TrustPass ID**. Repeat with a serial range off a product line and the endpoint enumerates the register.

The identifier is the part that matters. ADR 0004 spends its whole Context on the ID space not being enumerable — 128 bits of Crockford base32 with a check symbol, precisely so identifiers cannot be walked. This hands them out for a guess an attacker already had: the serial printed on the outside of the object.

## Why it was built that way, and why that reason still holds

It is not a mistake. A client that timed out and retried needs to know the identifier its first attempt created, or it has enrolled a product it cannot address. Returning "already enrolled" with no identifier makes a retry unrecoverable.

So the answer is not to remove it. The answer is that **it should only be answered for somebody entitled to the answer**, which today is nobody, because nothing authenticates.

## What should change

- The identifier is returned when the caller can be shown to be the one who enrolled it — retry recovery — and not otherwise.
- Rate limiting per actor and per address, since guessing is cheap and the reply is small.
- A rejection that does not distinguish "already enrolled by you" from "already enrolled by someone else" to an unauthenticated caller, because that distinction is the enumeration.

## Depends on

**#141 (TP-176), for one criterion only.** Corrected after reading it: this issue used to say it depended on `TP-141`, which resolves to **#102, closed** — the ADR *"who may speak, and why that authority was still valid when they spoke"*. `docs/ROADMAP.md` uses the same identifier for the authentication that was never filed, so a reader following the dependency found a closed issue and concluded this one was unblocked.

It is not, and it is also not blocked in the way that reading suggested. See the criteria below: **three of the four need no authentication at all**, because closing an information leak is removing something rather than adding a caller identity.

This is also why `docs/ROADMAP.md` says the API must not be exposed publicly yet — and this issue is a concrete instance of that warning rather than a new one.

## Acceptance criteria

**Buildable now, without authentication:**

- [ ] **The 409 stops naming the TrustPass ID.** A serial printed on the outside of an object stops being a lookup key for the identifier ADR 0004 spent its whole Context keeping unguessable. This is the disclosure, and it is the part that closes completely.
[...]

PR #125, opened 2026-09-18
[SECURITY] TP-161: the application connects as a role that owns nothing

### LINKED ISSUE #119: [SECURITY] TP-161: the API connects as a superuser that owns every table
## Measured

```
trustpass: superuser=true createdb=true
trustpass owns 8 of 8 public tables
```

That is the role in `DATABASE_URL`. The process serving HTTP can drop any trigger, alter any constraint, truncate any table and rewrite any history, because it owns them.

## Why this is the first security issue and not a detail of infrastructure

Every guarantee this project has built lives in Postgres, deliberately. `CONTRIBUTING.md` and six migrations say the same thing in different words: a rule that lives only in TypeScript is bypassed by the first path that writes without going through it, and `0005` chose a trigger over a `REVOKE` precisely because *"the application connects as the owner, and the owner can re-grant itself any privilege it was denied."*

That sentence was written as an argument **for** triggers. Read again, it is also a description of the hole:

```
API compromised
   ↓
superuser connection
   ↓
DROP TRIGGER lifecycle_event_no_update
   ↓
history is editable, and nothing records that it became so
```

The triggers are not a weaker defence than they look. They are exactly as strong as they look **and the credential can remove them**.

## Evidence that separation works

The #118 rehearsal ran seven attacks as a role with `SELECT`, `INSERT`, `UPDATE` and nothing else. All seven refused:

```
refused  edit history                 A lifecycle event cannot be modified…
refused  delete history               permission denied
refused  edit a grant                 A grant cannot be modified…
refused  drop the history trigger     must be owner of relation
refused  drop a table                 must be owner of table
refused  orphan a product's party     violates check constraint
refused  move a status with no event  A product cannot move from "registered"…
```

Two of those — dropping a trigger and dropping a table — refused **because of ownership**, and would succeed today with the credential the API actually uses.

## What should exist

```
owner role        owns the schema, used by nothing that serves traffic
migration role    DDL, used by db:migrate and nothing else
runtime role      SELECT / INSERT / UPDATE on what it needs, and no more
```

Raised in review, with a point worth keeping: `DROP TRIGGER` is not the only way back in. `ALTER TABLE`, `CREATE FUNCTION`, `CREATE TRIGGER`, `DROP FUNCTION` and `DROP INDEX` each let a compromised runtime rebuild or remove a defence indirectly.

## Acceptance criteria

- [ ] Three roles, with the runtime role holding no DDL privilege of any kind.
- [ ] `DATABASE_URL` for the API uses the runtime role; migrations use their own.
- [ ] The attacker suite extended beyond the seven above: `ALTER TABLE`, `CREATE FUNCTION`, `CREATE TRIGGER`, `DROP FUNCTION`, `DROP INDEX`, `TRUNCATE`, and `ALTER TABLE … DISABLE TRIGGER` — that last one especially, because it needs no `DROP` and leaves the trigger in `pg_trigger` looking present.
- [ ] Those run in CI, not once by hand.
- [ ] The runtime role's grants enumerated explicitly rather than `ALL TABLES`, so a new table is not writable by accident.
- [ ] Documented where the other database rules live.

## Not in scope

Authentication. This is about what the process may do once it has connected, which is a different question from who is asking — and it is worth fixing first, because it is the one that makes every other guarantee conditional.

## Risk

`critical`. It does not make the system wrong today. It makes every structural guarantee revocable by a single compromised credential, and the guarantees are the product.

clevervi:
Closes #119

## What this changes

Three roles. The application is never the owner of anything.

```
trustpass_owner      owns every object            NOLOGIN
trustpass_migration  DDL, by membership in owner  LOGIN, deploy only
trustpass_runtime    the application              LOGIN, no DDL, enumerated
```

`trustpass` stays superuser and stays the operator's way in. The boundary drawn
here is between the *application* and the record, which is the boundary an
attacker crosses over HTTP. An operator with the superuser password is a
different threat and #122's answer.

## Why, measured rather than assumed

`0005` argued for putting the guarantees in triggers, in a sentence that is also
the description of a hole:

> the application connects as the owner, and the owner can grant itself any
> privilege back.

Before deciding anything, against the running database:

```
trustpass: superuser=t createdb=t createrole=t replication=t bypassrls=t
trustpass owns 8 of 8 tables, 8 of 8 sequences, 7 of 7 functions
PUBLIC holds CONNECT and TEMP on the database (default, never revoked)
```

**Five cluster-level attributes, not the two the issue recorded.** `replication`
alone streams the entire cluster out without reading a table through SQL.

The `0023` rehearsal's seven refused attacks proved something about a role
invented for the rehearsal. Two of them — dropping a trigger, dropping a table —
succeed with the credential the API actually uses. And the cheapest attack was
not among the seven:

```sql
ALTER TABLE lifecycle_event DISABLE TRIGGER lifecycle_event_no_update;
```

One statement, no `DROP`, and the trigger stays listed in `pg_trigger` looking
entirely present. Only `tgenabled` moved.

## The matrix, and why it is this small

`ALL TABLES` has the defect `0023` refused in `CASCADE`: it acts without naming,
so a reviewer cannot say what it did. It is also wrong. Across every repository
and service:

```
.insert()  ->  product, lifecycle_event.  Nothing else.
.update()  ->  product.                   Nothing else.
.delete()  ->  nothing at all, anywhere.
```

| Table | Runtime |
| --- | --- |
| `product` | SELECT, INSERT, UPDATE |
| `lifecycle_event` | SELECT, INSERT |
| everything else | SELECT |
| any table | never DELETE, TRUNCATE, REFERENCES, TRIGGER |

`lifecycle_event` is the row that matters. The append-only trigger already
refuses an UPDATE with TP002, so withholding the privilege looks redundant — and
that is the idea. The trigger is what an attacker who reaches the owner turns
off. The missing privilege is what still says no afterwards, and it says it
*before* any trigger is consulted, because Postgres checks the privilege at
execution start. **The same attack #78 answered with TP002 is now answered with
`42501`, from a layer the layer below cannot disable.**

No sequence privileges either. Every `id` is `GENERATED ALWAYS AS IDENTITY`, not
`serial`, and an identity column's sequence advances without consulting the
caller. The rehearsal granted `USAGE, SELECT ON ALL SEQUENCES` for a need that
does not exist; the suite inserts a product as the runtime to prove it.

## Mutation table

Every mutation is a database change, applied, measured, reverted.

| --- | --- |
| runtime gains DELETE on history | `holds exactly lifecycle_event` |
| runtime gains UPDATE on history | `refused an edit to history before the trigger` |
| runtime gains TRUNCATE | `cannot truncate history` |
| runtime becomes a member of the owner | `cannot become the owner` |
| runtime gains CREATE on the schema | `cannot create anything in the schema` |
| runtime gains TEMP on the database | `cannot create anything in the schema` |
| runtime gains `replication` | `carries no cluster-level attribute` |
| runtime gains `bypassrls` | `carries no cluster-level attribute` |
| runtime is handed a table to own | `owns nothing` |
| a new table appears with no matrix entry | `covers every table in the schema` |
| functions go back to EXECUTE for PUBLIC | `can execute no function` |
| a SECURITY DEFINER function is added | `can execute no function` |
| the runtime is handed ALTER by ownership | `cannot alter a table` |

13 of 13 caught. `pnpm test`: **516 passed** (121 web, 335 db, 60 api), up from
485.

## Two findings from writing the suite

**A `GRANT` the runtime is not entitled to make does not raise.** Postgres emits
`WARNING: no privileges were granted` and returns successfully. Written like its
neighbours, that test failed with "the operation succeeded" and read like an
escalation — the catalogue said `DELETE = false`. An attacker checking exit
codes records this as a win; a defender checking exit codes records a breach.
The test asserts the outcome now, not the error.

**An `ALTER TABLE ... OWNER TO` destroys the grants held by the role becoming
the owner.** Moving `product` to the runtime and straight back left its ACL as
`{trustpass_owner=arwdDxtm/trustpass_owner}` — no runtime entry at all, no error,
no warning, while `lifecycle_event` beside it kept its own. In production every
read of the central table would have started returning `42501`. Recorded in the
migration, the ADR and `CONTRIBUTING.md`; nothing prevents it, the audit catches
it after the fact.

## What this does NOT do, and please read this part

**Nothing here takes effect until `DATABASE_URL` names `trustpass_runtime`.** A
migration cannot edit an environment file. Until an operator makes that change
this is a privilege model with nobody standing in it, and
`pnpm db:provision` says exactly that in its own output rather than letting
anyone believe otherwise.

It also does not separate `migration` from `owner` in *privilege*, only in
credential. Postgres lets only an owner alter an object, so a migration role
must be able to become the owner, and a member can `SET ROLE` at will —
compromising `trustpass_migration` is compromising `trustpass_owner`. ADR 0013
says so rather than claiming a blast radius the design does not have. What the
split buys: the owner has no password and no way to log in, and DDL has a
credential a deploy holds and the application never does.

## Review bar

Self-reviewed under `CONTRIBUTING.md`'s documented bar — one maintainer, no
second reader, and nothing here pretends otherwise. Mutations listed above,
findings posted inline.


packages/db/src/schema/least-privilege.integration.test.ts
```diff
@@ -0,0 +1,351 @@
...
+ * default somebody inherits.
+ */
+const EXPECTED: Record<string, readonly Grant[]> = {
+  actor: ["SELECT"],
+  capacity_grant: ["SELECT"],
+  capacity_grant_revocation: ["SELECT"],
+  credential: ["SELECT"],
+  lifecycle_event: ["SELECT", "INSERT"],
+  membership: ["SELECT"],
+  organization: ["SELECT"],
```
clevervi, line 54: **The line most likely to be wrong first.**

`organization` is SELECT-only because no write path reaches it today — verified across every repository, not assumed. But #120's fix moves enrolment toward creating an organization for an unknown party, and TP-141 brings issuer onboarding. Both need INSERT here.

That is the design working, not a gap: the GRANT ships in the migration that adds the capability, and this line changes in the same pull request. Recording it so that when the suite fails with `permission denied for table organization`, the first reading is "the privilege was not decided yet" rather than "least privilege is broken".

packages/db/drizzle/0024_three_roles_one_database.sql
```diff
@@ -0,0 +1,340 @@
...
+BEGIN
+  -- Every migration session becomes the owner on connect. Without this, a table
+  -- created by a future migration is owned by `trustpass_migration` instead, and
+  -- the ownership assertion below quietly stops being true for new objects —
+  -- the drift is invisible until something needs to alter one.
+  EXECUTE format(
+    'ALTER ROLE trustpass_migration IN DATABASE %I SET role = trustpass_owner',
+    current_database()
+  );
+END;
```
clevervi, line 124: **This assumes migrations will one day run as `trustpass_migration`, and today they do not.**

`SET role = trustpass_owner` only fires for sessions that connect as the migration role. The local `pnpm db:migrate` and the CI job both connect as `trustpass` superuser, so a table created by migration `0025` will be owned by `trustpass` and `owns nothing` / the ownership postcondition will catch it.

That failure is the correct one — it says the deploy is using the wrong credential — but it will read as a mysterious test failure to whoever hits it first. Worth challenging: should `0025` onwards carry an explicit `ALTER TABLE ... OWNER TO trustpass_owner` regardless, rather than relying on who is connected?

.github/workflows/ci.yml
```diff
@@ -117,9 +117,57 @@ jobs:
...
+      # here to rotate and nothing to leak beyond it. Masked anyway, because a
+      # password in a public build log teaches the wrong habit.
+      - name: Provision the runtime role
+        run: |
+          RUNTIME_PASSWORD="$(openssl rand -hex 24)"
+          MIGRATION_PASSWORD="$(openssl rand -hex 24)"
+          echo "::add-mask::$RUNTIME_PASSWORD"
+          echo "::add-mask::$MIGRATION_PASSWORD"
+          TRUSTPASS_RUNTIME_PASSWORD="$RUNTIME_PASSWORD" \
+          TRUSTPASS_MIGRATION_PASSWORD="$MIGRATION_PASSWORD" \
```
clevervi, line 136: `if: always()` is deliberate: the matrix is most worth reading on a run that failed. But it means the summary can print the state *before* provisioning if an earlier step died, which would show `trustpass_runtime` with no login and no grants and look like a catastrophic regression.

Acceptable for now because the suite itself is what enforces the matrix and this is only a display. Flagging it rather than fixing it, because the fix — a guard on the provisioning step having succeeded — adds a condition to read for a report nobody acts on directly.

PR #127, opened 2026-09-18
[SECURITY] TP-166: the API refuses to start on a connection that can disarm it

### LINKED ISSUE #126: [SECURITY] TP-166: the API will start happily as a superuser, and #119 left that to whoever writes the deploy
Follow-up to #119 / #125, and the half of it that a migration could not do.

## The gap #125 left open, in its own words

> Nothing here takes effect until `DATABASE_URL` names `trustpass_runtime`. A migration cannot edit an environment file.

True, and stated honestly rather than hidden. But what it leaves behind is a guarantee whose enforcement is a human remembering one line:

```
deploy forgets to change DATABASE_URL
  -> the API boots as `trustpass`
  -> superuser, createdb, createrole, replication, bypassrls
  -> every guarantee in ADR 0013 is off, and nothing says so
```

There is no signal. The service starts, the health check passes, the passports resolve. The only difference is that the append-only history is one `ALTER TABLE ... DISABLE TRIGGER` away from being editable, and the first evidence would be a record that changed.

## Objective

The API refuses to start against a connection that can dismantle its own guarantees.

## Scope

- A check in `packages/db` beside `isDatabaseReachable` in `health.ts`, asked of Postgres rather than parsed out of the URL.
- Called from `apps/api/src/index.ts` before `serve()`.
- Not called by the test suites, which use `createApp` directly and legitimately connect as a superuser to manage fixtures.

## Acceptance criteria

- [ ] The check asks Postgres, not the connection string. A role name in a URL is a claim; `current_user` is the answer.
- [ ] It refuses on any of `rolsuper`, `rolcreatedb`, `rolcreaterole`, `rolreplication`, `rolbypassrls`.
- [ ] **It refuses when the connected role owns any object in `public`**, which is the one the five flags miss. A role with every attribute false that owns `lifecycle_event` can still drop its trigger — ADR 0013 calls ownership the load-bearing half.
- [ ] The refusal names what was wrong and what to do, and exits non-zero. It does not warn and continue.
- [ ] **The escape hatch is an explicit opt-in, never an environment default.** See below.
- [ ] Using the escape hatch is logged at every start, not only the first.
- [ ] Integration-tested both ways against real Postgres: a superuser connection is refused, a `trustpass_runtime` connection starts.
- [ ] Mutation-checked: each condition removed in turn, a named test goes red.

## The one design point worth arguing about

The obvious shape is what the review proposed:

```
NODE_ENV=production AND current_user is privileged -> fail
```

That has the defect it is meant to fix. `apps/api/src/env.ts` declares:

```ts
NODE_ENV: z.enum(["development", "test", "production"]).default("development")
```

So a deploy that forgets `NODE_ENV` — the same forgetfulness this issue exists to catch, one variable over — gets `development` and **skips the check entirely**. The guard would be disarmed by exactly the mistake it was built for.

Inverted instead: the check runs everywhere, always, and the only way past it is a variable somebody sets on purpose:

```
TRUSTPASS_ALLOW_PRIVILEGED_DATABASE=true
```

Nobody sets that in production by accident, and forgetting any variable now makes the process *stricter* rather than weaker. A local `pnpm dev` against the superuser sets it once; `pnpm db:provision` plus a runtime `DATABASE_URL` is the alternative and the better habit.

## Out of scope

Rotating or managing the credentials themselves, and anything about the operator superuser. Backups are #122.

clevervi:
Closes #126

## Why

#119 built a role that cannot remove the guarantees it runs under, and said honestly what it could not do:

> Nothing takes effect until `DATABASE_URL` names `trustpass_runtime`. A migration cannot edit an environment file.

Which leaves a guarantee whose enforcement is a person remembering one line. Forget it and the API comes up as `trustpass` — superuser, owner of all sixteen objects — with **no signal of any kind**. Health green, passports resolving, and the append-only history one `ALTER TABLE ... DISABLE TRIGGER` from editable. The first evidence would be a record that changed.

## Executed, not built

As the superuser — what a forgotten `DATABASE_URL` looks like:

```
Refusing to start against the database as "trustpass":
  - it is a superuser, so no privilege check applies to it at all
  - it can replicate, which streams the whole cluster without reading a table
  - it bypasses row-level security, silently, the day any is defined
  - it can create roles, and so can grant itself anything through a new one
  - it can create databases
  - it can become the owner of 16 object(s) in public, which is enough to
    disable every trigger that protects the record

The application must connect as a role that owns nothing. See ADR 0013.
  1. pnpm db:migrate          creates trustpass_runtime and its grants
  2. pnpm db:provision        gives it a password from the environment
  3. point DATABASE_URL at trustpass_runtime

To run against this connection anyway, set TRUSTPASS_ALLOW_PRIVILEGED_DATABASE=true.

exit: 1
```

As `trustpass_runtime`:

```
{"level":"info","message":"api.started","port":3099,"environment":"development",
 "version":"0.4.0","databaseRole":"trustpass_runtime"}

GET /health   -> {"status":"ok","checks":{"database":"up"}}
POST /enrolments -> {"trustpassId":"TP1-1HSMQ3EHXMNWWJCXNFXBXFSA2JW", ... "status":"registered"}
```

A real enrolment through the restricted role: INSERT on `product`, INSERT on `lifecycle_event`, and the deferred provenance check at COMMIT.

## The two decisions worth disagreeing with

**1. Not gated on `NODE_ENV`.** The review proposed `NODE_ENV=production AND privileged -> fail`. That has the defect it exists to fix — `apps/api/src/env.ts` declares:

```ts
NODE_ENV: z.enum(["development", "test", "production"]).default("development")
```

A deploy that forgot `NODE_ENV` — the same forgetfulness this guard is for, one variable over — would **skip the check entirely**. The guard would be disarmed by exactly the mistake it was built for.

Inverted: the check runs everywhere, and the only way past is `TRUSTPASS_ALLOW_PRIVILEGED_DATABASE=true`. Nobody sets that in production by accident, and forgetting any variable now makes the process stricter.

**2. The condition is not "owns nothing".** That is the obvious implementation, it was what the review proposed, and it is wrong. Measured across all three roles:

| role | attributes clear | owns directly | **can become owner** |
| --- | --- | --- | --- |
| `trustpass` | no | 0 | 16 |
| `trustpass_migration` | **yes** | **0** | **16** |
| `trustpass_owner` | yes | 16 | 16 |
| `trustpass_runtime` | yes | 0 | **0** |

`trustpass_migration` carries no attribute and owns nothing, and one `SET ROLE trustpass_owner` makes it the owner of everything. A five-flags-plus-ownership check starts the application on it.

The question is what the connection **can become**, not what it holds. `pg_has_role(current_user, relowner, 'MEMBER')` answers both in one clause, because a role is a member of itself.

## Mutation table

| --- | --- |
| superuser no longer refused | `refuses a connection that has superuser` |
| replication no longer refused | `... has replication` |
| bypassrls no longer refused | `... has bypassrls` |
| createrole no longer refused | `... has createrole` |
| createdb no longer refused | `... has createdb` |
| ownership no longer refused | `can become an owner even with every attribute clear` |
| **membership query degraded to direct ownership** | `refuses a role that owns nothing but is one SET ROLE from owning everything` |
| failures reported one at a time | `reports every reason at once` |
| the escape hatch stops working | `starts anyway when told to` |
| the escape hatch warns silently | `starts anyway when told to` |
| an unanswerable check read as fine | `refuses to treat an unanswered question as a clean answer` |

11 of 11 caught — **on the second pass.** The first run caught 10, and the miss was the fail-closed branch: replacing `throw` with a permissive "everything false" left all sixteen tests green. The one branch whose entire job is to fail closed was the one nothing could prove. It has a test now.

`pnpm test`: **533 passed** (121 web, 352 db, 60 api), up from 516. Lint, typecheck and build clean.

## Consequences for anyone running this locally

**`pnpm dev` will now refuse to start against the superuser.** That is the point, and it is also friction on day one. Either:

```bash
TRUSTPASS_RUNTIME_PASSWORD=... TRUSTPASS_MIGRATION_PASSWORD=... pnpm db:provision
# then point DATABASE_URL at trustpass_runtime
```

or set `TRUSTPASS_ALLOW_PRIVILEGED_DATABASE=true` and accept a warning on every start.

The test suites are unaffected: they use `createApp` directly and never go through `index.ts`, and they legitimately need a superuser to manage fixtures.

## Review bar

Self-reviewed under `CONTRIBUTING.md`'s documented bar — one maintainer, no second reader, and nothing here pretends otherwise. Mutations listed above, findings inline.


apps/api/src/env.ts
```diff
@@ -4,6 +4,24 @@ const envSchema = z.object({
...
+   * Deliberately not keyed on NODE_ENV. That field carries
+   * `.default("development")` two lines up, so a deploy that forgot to set it —
+   * the same forgetfulness the check exists to catch — would disarm the check
+   * instead of triggering it. An explicit opt-in inverts that: forgetting any
+   * variable now makes the process stricter.
+   *
+   * Only the literal "true" enables it. `z.coerce.boolean()` would read
+   * "false", "0" and "no" as true, which is the wrong way for this one to be
+   * wrong.
+   */
```
clevervi, line 20: **The escape hatch is the remaining hole, and it is asymmetric in a way worth exploiting.**

Anyone who sets `TRUSTPASS_ALLOW_PRIVILEGED_DATABASE=true` in production defeats everything above it, and nothing detects that — the warning goes to a log nobody is reading at deploy time.

The argument in this comment is that `NODE_ENV` must not *loosen* the check, because its default is `development` and forgetting it would disarm the guard. But the reverse direction does not have that problem: using `NODE_ENV === "production"` to **refuse the escape hatch** is safe under the same forgetfulness, because forgetting it lands on `development` and the hatch merely keeps working, which is today's behaviour.

So the strictly better shape is probably:

```
privileged connection + escape hatch + NODE_ENV=production -> still refuse
```

Not done here because it can lock someone out of their own production during an incident, and that is a call to make deliberately rather than as a rider on this PR. Flagging it rather than deciding it.

packages/db/src/connection-privileges.ts
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Asks Postgres, rather than reading the role out of the connection string.
+ *
+ * A role name in a URL is a claim about what was intended. `current_user` is
+ * what happened — and they differ whenever a connection pooler, a `SET ROLE` in
+ * a session default, or a plain typo sits between the two.
+ */
+export async function readConnectionPrivileges(db: Database): Promise<ConnectionPrivileges> {
+  const [row] = await db.execute<{
```
clevervi, line 100: **What this proves, precisely: one connection, once, at startup.**

`createDatabase` opens a pool of ten. This asks `current_user` on whichever member the query lands on, at boot, and nothing re-checks afterwards.

That is sound for the threat it addresses — a `DATABASE_URL` pointing at the wrong role means every connection in the pool is that role. It is *not* sound against a connection pooler sitting between the API and Postgres that hands out sessions belonging to different roles, which is exactly what PgBouncer in `session` mode with several server users would do.

Nothing like that exists here yet. Recording the boundary of the claim so nobody later reads this as "every query runs unprivileged" when it says "the connection the process opened at boot was unprivileged".

PR #129, opened 2026-09-18
[SECURITY] TP-167: the startup guard could not see a function it had to protect

### LINKED ISSUE #128: [SECURITY] TP-167: the startup guard is blind to function ownership, which is how the quiet attack works
A defect in #127, merged three hours ago, found by taking a review point seriously enough to test it.

## What is wrong

`readConnectionPrivileges` counts reachable ownership like this:

```sql
FROM pg_class
WHERE relnamespace = 'public'::regnamespace
  AND relkind IN ('r', 'S', 'v', 'm')
```

`pg_class` does not contain functions. Measured:

```
tables/sequences/views in public (what the guard counts)   16
functions in public (what it does NOT count)                7
objects in the drizzle schema (also not counted)            2
```

And demonstrated with a probe role that can become the owner of a function and nothing else:

```
shipped guard says                        0     <- the API would start
but it can redefine this many functions   1
```

## Why it matters, in this repository specifically

Migration `0024` transfers those seven functions on purpose, and says why:

> a function left owned by the old role is a function the old role can still redefine — `CREATE OR REPLACE FUNCTION` on an append-only guard is a quieter attack than dropping it, because the trigger stays listed in `pg_trigger` and keeps firing.

`least-privilege.integration.test.ts` has that exact attack in its table, named `redefine a trigger function`. So the migration protects function ownership, the runtime suite tests that the runtime cannot reach it, and **the startup guard does not look at it at all.** Three layers, one of them facing the wrong way.

The `drizzle` schema is the smaller half of the same mistake: a role that can become the owner of `__drizzle_migrations` can rewrite which migrations this database believes it has run.

## Scope

Extend the inventory to everything in a non-system schema that carries a guarantee:

| Catalogue | Why it belongs |
| --- | --- |
| `pg_class` (`r`, `p`, `S`, `v`, `m`) | the tables the record lives in |
| `pg_proc` | the seven trigger functions that enforce TP001–TP005 |
| `pg_type` (`e`, `d`, `r`) | `lifecycle_actor_kind` is an enum; altering it silently reinterprets every event recorded under a value. `0023` already has a test protecting one of its values |
| any schema but `pg_catalog`, `information_schema`, `pg_toast*`, `pg_temp*` | `drizzle.__drizzle_migrations` is outside `public` |

Measured with the corrected query:

```
role                  shipped   corrected
trustpass             16        35
trustpass_migration   16        25
trustpass_owner       16        25
trustpass_runtime     0         0      <- the working path is unaffected
```

## Acceptance criteria

- [ ] Function ownership counts. A role that can become the owner of a trigger function and nothing else is refused.
- [ ] Type ownership counts, for enums, domains and ranges. Auto-generated composite and array types are excluded, or every table is counted twice.
- [ ] Non-system schemas count, `drizzle` included.
- [ ] `trustpass_runtime` still starts. A guard that refuses everything is an outage, not a control.
- [ ] Mutation-checked: each catalogue removed from the union in turn, a named test goes red.
- [ ] The integration test builds the probe role rather than asserting against a number, so the test does not need updating every time a migration adds an object.

## And one correction to #127 that is not a bug

The review proposed `pg_has_role(..., 'SET')` instead of `'MEMBER'`, on the grounds that `SET` is what means "can `SET ROLE`". The reading of the documentation is right and **the change would open a hole**, measured on Postgres 18:

```
role               MEMBER   SET     USAGE (inherits privileges)
tp_probe_inherit   t        f       t
```

`GRANT trustpass_owner TO tp_probe_inherit WITH INHERIT TRUE, SET FALSE` produces a role that cannot `SET ROLE` and holds every owner privilege anyway, by inheritance. Connected as it, without ever issuing `SET ROLE`:

```
ALTER TABLE lifecycle_event DISABLE TRIGGER lifecycle_event_no_update;
-> ALTER TABLE
guard_on: 0   still_listed_in_pg_trigger: 1
```

A `SET`-based guard counts `0` for that role and starts the application on it. `MEMBER` is `USAGE OR SET` and catches both paths, which is the property wanted here.

What the review is right about is the prose: the comment in `connection-privileges.ts` says `MEMBER` answers "what it can become", and it does not. It answers "is a member of, directly or indirectly", and the reason that is the correct question is that **inheritance reaches the same privileges without `SET ROLE` ever being called**. The wording should say that.

clevervi:
Closes #128

A defect in #127, merged four hours ago. Found by testing a review objection rather than accepting or dismissing it — and the objection turned out to be right about a problem and wrong about the fix.

## The defect

The ownership inventory asked `pg_class` in `public` only:

```
tables/sequences/views in public (what it counted)   16
functions in public (what it did NOT count)           7
objects in the drizzle schema (also not counted)      2
```

`pg_class` does not contain functions. Demonstrated with a probe role able to become the owner of one function and nothing else:

```
shipped guard says                        0     <- the API would start
but it can redefine this many functions   1
```

### Why that is worse here than it sounds

Migration `0024` transfers those seven trigger functions to the owner on purpose, and says why:

> a function left owned by the old role is a function the old role can still redefine — `CREATE OR REPLACE FUNCTION` on an append-only guard is a quieter attack than dropping it, because the trigger stays listed in `pg_trigger` and keeps firing.

`least-privilege.integration.test.ts` carries that exact attack, named `redefine a trigger function`. So the migration protected function ownership, the runtime suite tested that the runtime cannot reach it, and **the startup guard did not look at it at all.** Three layers, one facing the wrong way.

## The fix

`pg_class` + `pg_proc` + `pg_type` across every non-system schema.

| Catalogue | Why |
| --- | --- |
| `pg_class` (`r`, `p`, `S`, `v`, `m`) | the tables the record lives in |
| `pg_proc` | the seven trigger functions enforcing TP001–TP005 |
| `pg_type` (`e`, `d`, `r`) | `lifecycle_actor_kind` is an enum; dropping a value silently reinterprets every event recorded under it, which `0023` already defends |
| non-system schemas | `drizzle.__drizzle_migrations` decides which migrations this database believes it has run |

Composite and array types are excluded, or every table counts three times.

```
role                  before   after
trustpass             16       35
trustpass_migration   16       25
trustpass_owner       16       25
trustpass_runtime     0        0      <- the working path is unaffected
```

## The review's other proposal, measured

It suggested `pg_has_role(..., 'SET')` instead of `'MEMBER'`, on the grounds that `SET` is the privilege that means "can issue `SET ROLE`". **The reading of the documentation is correct and the change opens a hole.** On Postgres 18:

```sql
GRANT trustpass_owner TO probe WITH INHERIT TRUE, SET FALSE;
```

| role | MEMBER | SET | USAGE (inherits privileges) |
| --- | --- | --- | --- |
| `tp_probe_inherit` | t | **f** | **t** |

Connected as that role, never issuing `SET ROLE`:

```
ALTER TABLE lifecycle_event DISABLE TRIGGER lifecycle_event_no_update;
-> ALTER TABLE
guard_on: 0    still_listed_in_pg_trigger: 1
```

A `SET`-based count reads **0** for it and starts the application on a connection holding every owner privilege by inheritance. `MEMBER` is `USAGE OR SET` and covers both paths.

Where the review is right: the prose. The comment said `MEMBER` answers "what it can become", and it does not — it answers "is a member of, directly or indirectly", and the reason that is the correct question is that **inheritance reaches the same privileges without `SET ROLE` ever being called**. The wording says that now, and so does ADR 0013.

## Mutation table

| --- | --- |
| functions drop out of the inventory | `sees a function it could redefine, which pg_class does not contain` |
| enums and domains drop out | `sees an enum it could reinterpret every past event with` |
| the inventory narrows back to `public` | `sees an object outside public, where the migration ledger lives` |
| **`MEMBER` becomes `SET`, as review proposed** | `sees a member that inherits the owner's rights but cannot SET ROLE` |

4 of 4. That last row is the one worth noting: before this PR the reasoning against `SET` existed only in a comment, and **a comment cannot go red**.

The new tests build their own probe role rather than assert a count. The count was 16 last week and is 25 today; a test pinned to either fails on the next migration for no reason anyone cares about.

`pnpm test`: **537 passed** (121 web, 356 db, 60 api). Lint, typecheck, build clean.

## Scope added deliberately

One thing here is not in #128: a CI step that fails if `TRUSTPASS_ALLOW_PRIVILEGED_DATABASE` is set. It answers the review's point about the escape hatch turning fail-closed into fail-open, and it is the part of that policy CI can actually enforce — a run with the hatch set would never exercise the boundary the suite exists to test. **It cannot speak for production**, which is why the rest of the rule belongs in deployment policy and is named as such rather than implied.

## Review bar


packages/db/src/connection-privileges.ts
```diff
@@ -106,16 +124,52 @@ export async function readConnectionPrivileges(db: Database): Promise<Connection
     bypass_rls: boolean;
     reachable_ownership: string;
   }>(sql`
+    WITH ns AS (
+      -- Every schema this database's own objects can live in. The first
+      -- version asked about 'public' alone and missed
+      -- drizzle.__drizzle_migrations, whose owner decides which migrations
+      -- this database believes it has already run.
+      SELECT oid FROM pg_namespace
+      WHERE nspname NOT IN ('pg_catalog', 'information_schema')
```
clevervi, line 133: **The schema filter is now "everything that is not Postgres", which is broader than it was and could produce a false refusal.**

An extension installed into its own schema — PostGIS, `pg_stat_statements`, anything added later — puts objects here. They are normally owned by a superuser, so `trustpass_runtime` is not a member and the count stays 0, which is why this is a note and not a change.

Where it would bite: an extension installed by a role the runtime is a member of, or a schema created for a reporting tool whose objects the runtime legitimately owns. The refusal would be correct in the first case and an outage in the second.

The alternative was an allowlist of schemas, which fails the other way — it is the shape that produced this very defect, since the first version allowlisted `public` and missed `drizzle`. A filter that is too wide refuses loudly; one that is too narrow starts the application. Choosing the loud failure deliberately.

PR #131, opened 2026-09-18
[SECURITY] TP-168: two claims in #129 were false, and one hid a real attack

### LINKED ISSUE #130: [SECURITY] TP-168: two claims in #129 are false, and the attack behind one of them is untested
Review returned two precision corrections on #129. Both tested, both correct, and the second one led to a real gap.

## 1. "`MEMBER` is `USAGE OR SET`" is false

Written in `connection-privileges.ts` and in ADR 0013 to justify choosing `MEMBER`. It is not an equivalence. Measured on Postgres 18.6:

```sql
GRANT tp_m_owner TO tp_m_none WITH INHERIT FALSE, SET FALSE;
```

| grant shape | MEMBER | USAGE | SET |
| --- | --- | --- | --- |
| `INHERIT FALSE, SET FALSE` | **t** | f | f |

`MEMBER` is strictly broader: it reports membership regardless of what that membership confers. So the guard can refuse a connection whose membership grants it nothing at all.

**That does not make the choice wrong** — a false positive in a fail-closed barrier is the direction to be wrong in, and it means no membership shape can slip past the inventory. But the reason has to be "deliberately conservative", not an equivalence that does not hold. Claiming a property the catalogue does not support is the same defect as a stale README.

## 2. The enum justification is wrong, and the real attack is worse

`connection-privileges.ts` and its test say a type owner can drop an enum value and silently reinterpret history. Postgres 18.6:

```
ALTER TYPE tp_enum_probe DROP VALUE 'c';
ERROR:  dropping an enum value is not implemented
```

So that attack does not exist. What does exist is `ALTER TYPE ... RENAME VALUE`, and it is worse. Run against the real type, inside a transaction, rolled back:

```
before                      after ALTER TYPE lifecycle_actor_kind
actor_kind | count          RENAME VALUE 'issuer' TO 'holder_verified'
-----------+-------         ---------------+-------
issuer     |  1744          holder_verified|  1744
authority  |  1167          authority      |  1167
holder     |   722          holder         |   722
```

**1,744 lifecycle events restated, with zero UPDATEs.** `lifecycle_event_no_update` never fires, because no row is touched. The append-only guarantee — TP002, ADR 0005, the whole argument of #78 — is bypassed by editing the dictionary rather than the text.

Keeping `pg_type` in the inventory was the right call. The justification was not.

## 3. And nothing tests that the runtime cannot do it

`least-privilege.integration.test.ts` carries 13 attacks. None of them is `ALTER TYPE`. The runtime is presumably refused — it owns nothing — but presumably is what this repository keeps replacing with a measurement.

## Acceptance criteria

- [ ] `ALTER TYPE lifecycle_actor_kind RENAME VALUE` is in the runtime attack table and refused with `42501`.
- [ ] The `MEMBER` justification says "deliberately conservative: any direct or indirect membership, whatever it confers" and names the false-positive it accepts, in the module and in ADR 0013.
- [ ] The `pg_type` justification names `RENAME VALUE` and `DROP TYPE`, with the measurement, and stops claiming a `DROP VALUE` that Postgres does not implement.
- [ ] Mutation-checked: the new attack case dies if the runtime is granted what it needs.

## 4. And one boundary that belongs in the code rather than a PR comment

Raised in review, and already said in an inline comment on #127 where nobody will find it again: **the guard is a startup barrier, not a continuous guarantee.** It asks once, on one pooled connection, before serving. A membership changed at 10:20 against a process that started at 10:05 is not detected.

The property it actually provides is *"the connection was in the right posture when the application started"*, and that is a defence against a misconfigured deployment rather than against a compromised administrator. Per-request privilege checks are explicitly **not** wanted — they would cost a round trip per request to detect something the role model already prevents.

- [ ] Stated in `connection-privileges.ts` and in ADR 0013's Consequences, in those terms.

## Not in scope

Re-checking privileges during the process lifetime. See above.

clevervi:
Closes #130

Review returned two precision corrections on #129. Both tested, both correct, and the second one led somewhere.

## 1. `MEMBER is USAGE OR SET` is false

Written in `connection-privileges.ts` and ADR 0013 to justify choosing `MEMBER`. Measured on Postgres 18.6:

```sql
GRANT tp_m_owner TO tp_m_none WITH INHERIT FALSE, SET FALSE;
```

| grant shape | MEMBER | USAGE | SET |
| --- | --- | --- | --- |
| `INHERIT FALSE, SET FALSE` | **t** | f | f |

`MEMBER` is strictly broader: it reports membership whatever that membership confers. So the guard can refuse a connection whose membership grants it nothing at all.

**The choice stays; the reasoning changes.** A false positive in a fail-closed barrier is the direction to be wrong in, and it means no future shape of `GRANT ... WITH` slips past the inventory. Conservative, not exact — and claiming a property the catalogue does not support is the same defect as a stale README.

## 2. The enum justification was wrong, and the real attack is worse

The comment said a type owner can drop an enum value. Postgres 18.6:

```
ALTER TYPE tp_enum_probe DROP VALUE 'c';
ERROR:  dropping an enum value is not implemented
```

`RENAME VALUE` exists. Run against the real type inside a rolled-back transaction:

```
ALTER TYPE lifecycle_actor_kind RENAME VALUE 'issuer' TO 'holder_verified';

before                    after
issuer      1744          holder_verified  1744
authority   1167          authority        1167
holder       722          holder            722
```

**1,744 lifecycle events restated, with zero UPDATEs.** `lifecycle_event_no_update` never fires, because no row is touched. The append-only guarantee ADR 0005 established and #78 defended is bypassed by editing the dictionary instead of the text — and nothing in TP001–TP005 sees it either.

### And nothing tested that the runtime cannot do it

`least-privilege.integration.test.ts` carried thirteen attacks and not one `ALTER TYPE`. It has fourteen now.

| --- | --- |
| `ALTER TYPE lifecycle_actor_kind OWNER TO trustpass_runtime` | `rename an enum value out from under every event recorded with it` |

The statement names a value that does not exist, deliberately. Postgres checks ownership before it looks the value up — `42501: must be owner of type lifecycle_actor_kind` — so the case still tests the boundary and stays **harmless if the boundary ever goes**, failing on the missing value instead of renaming a live one.

That change was not foresight. The first version renamed `'issuer'` for real during the mutation check and left 1,744 events reading `tp_probe_renamed` until it was put back. An attack test that does damage when it succeeds is a test nobody should run near anything that matters.

## 3. A race #129 introduced, found on the way

The full suite failed once while the two files involved passed together every time — **order-dependent, not flaky**, the distinction #78 already cost this project a morning.

#129's probes created a table in `public` and a function with the default ACL, which is `EXECUTE` to `PUBLIC`. Vitest runs test files in parallel, and `least-privilege` asserts two things over `public` at the same moment: the exact list of tables, and that the runtime can execute no function.

```
full suite, before:  1 failure in 4 runs
full suite, after:   0 failures in 8 runs
```

Every probe is now a function with `EXECUTE` revoked in the same statement that creates it, or lives in its own schema.

## 4. The boundary, moved out of a PR comment

Raised in review, and already written on #127 where nobody will find it again: **the guard is a startup barrier, not a continuous guarantee.** It asks once, on one pooled connection, before serving. A membership granted at 10:20 against a process started at 10:05 is not detected.

The property is *"the connection was in the right posture when the application started"* — a defence against a misconfigured deployment, not a compromised administrator. Per-request checks were considered and **rejected**: a round trip on every request to detect what the role model already prevents. Now in the module and in ADR 0013's Consequences.

## Also verified

The review reported that #122 still showed only its original four acceptance criteria. It does not — the roles/catalogue/artefact/data criteria are published, `updatedAt 2026-09-18T18:30:36Z`, four marker matches in the body. A cached view.

`pnpm test`: **538 passed** (121 web, 357 db, 60 api). Lint, typecheck, build clean. 4/4 inventory mutations still caught after the probe objects changed shape.

## Review bar


packages/db/src/schema/least-privilege.integration.test.ts
```diff
@@ -114,6 +114,36 @@ const ATTACKS: readonly { name: string; statement: string }[] = [
...
+    // lifecycle_actor_kind`. So this asserts the privilege and stays harmless
+    // the day the privilege is gone — it would fail on the missing value
+    // instead of renaming a live one.
+    //
+    // Written that way after the mutation check renamed `'issuer'` for real and
+    // left 1,744 events reading `tp_probe_renamed` until it was put back. An
+    // attack test that does damage when it succeeds is a test nobody should run
+    // near anything that matters.
+    name: "rename an enum value out from under every event recorded with it",
+    statement:
```
clevervi, line 144: **This case is weaker than the thirteen around it, and the trade is deliberate.**

Every other attack in this table is the real statement. This one names `tp_probe_absent`, a value that does not exist, and relies on Postgres checking ownership before it looks the value up.

What it costs: if that ordering ever changes, the case starts failing on the wrong error and reads as a broken test rather than a broken boundary. Verified today — `42501: must be owner of type lifecycle_actor_kind` — but it is an assumption about statement internals, not about privileges, and nothing here would notice it drifting.

What it buys: the suite stops being able to rename a live enum value. The mutation check did exactly that, and 1,744 events read `tp_probe_renamed` until it was put back by hand. A test that can restate history when it fails does not belong anywhere near a database that matters, and the other thirteen cases are safe because their statements are refused *and* individually harmless — `DROP TABLE organization` against a role that could do it would be caught by everything else long before this file ran.

Worth challenging: an alternative is asserting the ownership from the catalogue instead of attempting the statement, which has no ordering assumption and no blast radius — but it stops being an attack test and becomes another row of the grant matrix, which is a different thing and already covered.

PR #133, opened 2026-09-18
[CORE] TP-169: an AI context router, so reading a large file does not cost a large file

### LINKED ISSUE #132: [CORE] TP-169: an AI context router, so reading a large file does not cost a large file
Development tooling in `.claude/`. It does not touch `apps/api`, `apps/web` or `packages/db`, and nothing about the product depends on it.

## The problem

Reading a 1,500-line file to find out what is in it costs the whole file, every time, and most of it is never needed. A cheaper model with a large context window can read it once and hand back a map.

## The rule that makes this safe

**Claude is authoritative. Gemini is a bulk reader that is allowed to be absent.**

```
Gemini has no key / 401 / 403 / 429 / quota / timeout / 5xx / bad JSON / bad schema
        -> record the reason
        -> open the circuit
        -> Claude reads the file directly
        -> never a retry loop
```

Killing Gemini must not slow the work down, and one of the tests proves it by unsetting the key.

## Risk, and why this is `risk:high` for a tooling change

A defect here sends source code to a third party. Google states that Free Tier content may be used to improve their products, so the classification is not a nicety:

| Level | Meaning | Examples |
| --- | --- | --- |
| **S0** | never transmitted, **hardcoded, not from a config file** | `.env*`, `*.pem`, `*.key`, `*.p12`, `*.pfx`, credentials, tokens, SSH and cloud material |
| **S1** | Claude only | `packages/db/drizzle/**`, triggers, auth, authorization, crypto, ADRs, lifecycle invariants |
| **S2** | candidate | API, services, components |
| **S3** | preferred | large tests, docs, repetitive source |

**S0 is in code, not in `sensitive-paths.txt`.** A denylist the repository can edit is a denylist a pull request can empty, and this project already refused that shape of argument in #119: a guard the attacker can disable is not a guard. The policy file may *add* rules and may never remove an S0 one, and if it is missing, unreadable or malformed the router fails closed towards Gemini while Claude carries on.

**Paths are resolved through symlinks before classification.** `safe/foo.ts` pointing at `.env` must not be readable as safe.

**Files outside the repository are never delegated.**

## Verified before writing any of it

| Claim | Result |
| --- | --- |
| `PreToolUse` can deny a `Read` | Yes — `hookSpecificOutput.permissionDecision: "deny"` |
| The deny reason reaches Claude | **Yes.** Without this the whole design is impossible |
| Payload field names | `tool_name`, `tool_input.file_path` — snake_case |
| A hook that times out | Does **not** block; the Read proceeds. Fail-open by default |
| `node script.mjs` on Windows | Works; exec form behaves identically to Unix |
| `x-goog-api-key` header | Correct, and the key never goes in the URL |
| `gemini-3.8-flash` | Real, current |
| `generationConfig.responseMimeType` + `responseSchema` | Correct for `generateContent` |
| `thinkingLevel: "low"` | **Unverified.** Docs show `minimal` for 3.8. Not sent by default; see below |

## Acceptance criteria

- [ ] `PreToolUse` hook on `Read`. Malformed or unexpected stdin exits 0 and changes nothing.
- [ ] A bounded read (`limit` <= 180) is never delegated, so Claude can verify exact lines after reading a summary.
- [ ] A large eligible read produces a local context file and denies the original read, naming the file, its SHA-256, and that the summary is navigation rather than authority.
- [ ] **Cache is consulted before quota.** A valid summary needs no Gemini call, and works while the circuit is open.
- [ ] Cache key covers source SHA-256, model, prompt version, router version and extractor version.
- [ ] A deterministic local extractor runs first, and Gemini is skipped when it is enough.
- [ ] Structured output enforced by JSON Schema **and** validated again locally — line ranges inside the file, symbols that exist, the source hash echoed back.
- [ ] Gemini output is treated as untrusted data. It never becomes a command, a patch, an edit or a tool call.
- [ ] The source is untrusted too: instructions inside a comment or a string are never instructions.
- [ ] Circuit breaker: quota and auth failures disable immediately with no retry; timeouts and 5xx get **one** retry.
- [ ] State written under a filesystem lock, so two processes cannot interleave.
- [ ] No secret appears in any log, cache entry or state file. The key lives in process memory only.
- [ ] `node .claude/scripts/tp-stats.mjs` reports reads, delegations, cache hits, failures, fallbacks and an **estimated** token saving, labelled as an estimate.
- [ ] 21 tests under `node --test`, including: the key absent, a 429, one timeout then a second, malformed JSON, a symlink to a secret, a file outside the repo, prompt injection in the source treated as data, and a repeated full read not re-calling Gemini.
- [ ] `pnpm lint && pnpm typecheck && pnpm test && pnpm build` still clean.

## Two decisions this issue makes rather than defers

**`.gitignore` currently ignores all of `.claude/`.** That changes: the router source is tracked, and `cache/` and `.state/` stay ignored. A hook deciding what leaves this machine is a security control, and this repository's rule is that security controls are written down, reviewed and tested. An untracked one is the "trust me" #119 spent a day deleting.

**`thinkingLevel` is not sent by default.** The review asked for `low`; the documentation for 3.8 shows `minimal`. Neither could be verified without a live call, and a wrong value is a 400 on every request. It is settable through `TP_GEMINI_THINKING_LEVEL`, and the default costs latency on Gemini's side rather than correctness on ours.

## Explicitly not in scope

Write or Edit delegation. An MCP server for Gemini. A Gemini agent. Anything that lets it produce a patch. This pull request is read optimisation, cache, fallback and metrics.

clevervi:
Closes #132

Development tooling in `.claude/`. It does not touch `apps/api`, `apps/web` or `packages/db`, and nothing about the product depends on it.

## What it does

A `PreToolUse` hook on `Read`. An eligible large file is replaced with a local map and the original read is denied, so Claude receives a few hundred tokens instead of a few thousand and then asks for the exact lines it needs.

```
Read(file) -> hook -> S0/S1? outside the repo? bounded read? small file? -> pass through
                   -> otherwise: hash -> cache -> local scanner -> Gemini -> map -> deny
```

**Claude is authoritative. Gemini is a bulk reader that is allowed to be absent.**

## Verified before any of it was written

The whole design rests on four facts, and guessing any of them would have produced a tool that does not work:

| Claim | Result |
| --- | --- |
| `PreToolUse` can deny a `Read` | Yes — `hookSpecificOutput.permissionDecision: "deny"` |
| **The deny reason reaches Claude** | **Yes.** Without this the design is impossible |
| A hook that times out | Does **not** block; the read proceeds. Fail-open is the default |
| `x-goog-api-key` header, key never in the URL | Correct, and a test asserts the URL contains neither the key nor `key=` |
| `gemini-3.8-flash`, `responseMimeType` + `responseSchema` | Correct for `generateContent` |
| `thinkingLevel: "low"` | **Could not be verified.** The docs for this model show `minimal`. Not sent by default — see below |

## The threshold was measured, not chosen

It was 600 lines first. Counted against the repository it runs in:

```
>= 200 lines   30 files
>= 300 lines   12 files
>= 400 lines    7 files
>= 500 lines    3 files
>= 600 lines    0 files
```

**At 600 this router would never once have fired here.** At 400 it has seven files to work with. That is a small number and it is the honest one: on a codebase this size the mechanism matters more than the saving, and the saving arrives when the codebase does. Tuning the threshold until the tool looked useful was the other option.

## Security, which is why this is `risk:high` for a tooling change

A defect here sends source code to a third party, and Google states Free Tier content may be used to improve their products.

**S0 is hardcoded and is not read from `sensitive-paths.txt`.** `.env*`, keys, certificates, `credentials`, `secrets`, `.git/`. The policy file can *add* a rule and can never remove one; when it is missing or malformed, **nothing** is delegated. The argument is #119's: a guard the attacker can switch off is not a guard, and a denylist living in the repository is one pull request from being a shorter denylist.

**Paths are resolved through symlinks before classification**, because `docs/harmless.md` pointing at `.env` is an `.env`. **Files outside the repository are never delegated.**

**Both directions are untrusted.** The source is data — a comment saying "ignore previous instructions" is program text being reported on, and the system instruction plus a fenced region say so. The output is data — the response schema has no field for an action, a command or a patch, because a field that does not exist cannot be filled in, and every line range is checked against the file that was actually sent.

## Failure policy

```
no key / 401 / 403 / 429 / budget spent   -> no retry, circuit opens, Claude reads the file
timeout / 5xx                             -> exactly one retry, then the same
bad JSON / bad schema / bad line ranges    -> no retry, circuit opens
```

**The cache is consulted before the budget and before the circuit**, so a summary already paid for does not need Gemini to be reachable.

## Tests

**55, and every one runs with Gemini switched off.** A suite that needed the optional dependency would be testing the wrong thing.

The integration tests run the real hook as a real process, payload on stdin, nothing stubbed, `GEMINI_API_KEY` empty — including one that demonstrates the entire flow with no network and no key, because a large test file is a list of similarly shaped things and the local line scanner is enough for that.

```
node --test ".claude/tests/*.test.mjs"
tests 55   pass 55   fail 0
```

`pnpm lint && pnpm typecheck && pnpm test && pnpm build` all clean; **538 project tests** unchanged. Biome lints `.claude/` too, so the router is held to the same bar as the product.

## Two defects the tests found

**The denial gave a repository-relative path.** It reads fine to a human and sends the `Read` tool looking in the wrong place. Caught only because the integration test puts the state directory elsewhere, where the relative form came out as `../../AppData/Local/Temp/...`.

**The first delegation target chosen was `schema/least-privilege.integration.test.ts`**, which the router correctly refuses — it is S1. The test was wrong, not the classifier.

## Decisions made rather than deferred

**`.gitignore` stops ignoring all of `.claude/`.** The router source is tracked; `cache/` and `.state/` are not. A hook that decides what leaves this machine is a security control, and the rule here is that security controls are written down, reviewed and tested — an untracked one is the "trust me" #119 spent a day deleting.

**`thinkingLevel` is not sent.** The review asked for `low`; the documentation for this model shows `minimal`. A wrong value is a 400 on every request, and the default costs latency on Gemini's side rather than correctness on ours. `TP_GEMINI_THINKING_LEVEL` sets it once somebody has confirmed which.

## Not in scope

Write or Edit delegation. An MCP server. A Gemini agent. Anything that lets it produce a patch.

## Review bar


.claude/scripts/tp-router.mjs
```diff
@@ -0,0 +1,283 @@
...
+  /(^|\/)(auth|authz|authorization|security|crypto|session|token)(\/|\.|-)/,
+  /(^|\/)connection-privileges/,
+  /(^|\/)least-privilege/,
+  /(^|\/)provision-roles/,
+  /^CONTRIBUTING\.md$/,
+  /^SECURITY\.md$/,
+];
+
+/** Bulk by nature. Reading one of these in full is usually waste. */
+const S3_PATTERNS = [/\.test\.(ts|tsx|mjs|js)$/, /\.spec\.(ts|tsx|mjs|js)$/, /(^|\/)fixtures?\//];
```
clevervi, line 81: **The honest limit of this whole classifier: it works on paths, and a path is not a fact about contents.**

`apps/web/app/page.tsx` is S2 because it is ordinary source. If somebody pastes a live token into it as a constant, it stays S2 and it goes to Google.

S0 catches the files whose *names* announce what they hold, which is most of the real cases and none of the interesting ones. A content scan before transmission — entropy, known token prefixes — would catch the rest, and is not here.

Not added because a content scanner that is wrong in the permissive direction reads as safety while providing none, and one that is wrong in the strict direction blocks ordinary source at random. It deserves its own decision rather than a rider on this. Recording the gap so it is a known one.

The mitigation that does exist: gitleaks already scans every commit and the full history weekly, so a token in source is caught at the repository boundary — just not before this hook would have sent it.

clevervi:
Self-review. Two gaps I want on the record rather than smoothed over.

## The classifier works on paths, and a path is not a fact about contents

`apps/web/app/page.tsx` is S2 because it is ordinary source. If somebody pastes a live token into it as a constant, it stays S2 and it goes to Google.

S0 catches the files whose **names** announce what they hold. That is most of the real cases and none of the interesting ones. A content scan before transmission — entropy, known token prefixes — would catch the rest, and is not here.

Not added because a content scanner that errs permissive reads as safety while providing none, and one that errs strict blocks ordinary source at random. It deserves its own decision rather than a rider on this pull request.

What does exist: gitleaks scans every commit and the full history weekly, so a token in source is caught at the repository boundary — just not before this hook would have sent it.

## `extractionIsSufficient` is the crudest thing here, deliberately

S3 plus ten symbols, and nothing else. A 500-line test file never reaches Gemini; a 500-line component always does, with no consideration of whether it was structured enough to map locally.

The cleverer version — scoring how much of the file the symbol map covers — was written and thrown away. Its failure mode is deciding a subtle file is "structured enough", and what that produces is a confident summary of something nobody read. This rule fails the other way: it sends files that the scanner might have handled, which costs a call and nothing else.

Worth revisiting once there is data on how often the local map was enough. There is none today, and inventing a heuristic before the data is exactly how the 600-line threshold got picked in the first place.

## And one thing that is not a gap

The 55 tests did not run in CI when this PR was opened — `pnpm test` does not reach `node --test`. Fixed in the last commit, and confirmed in the run: `tests 55 / pass 55`. A suite that only runs where somebody remembered to run it is the same class of claim this project keeps replacing with a check.

PR #138, opened 2026-09-18
[OPERATIONS] TP-164: a backup that was restored, and the two files it needed

### LINKED ISSUE #122: [OPERATIONS] TP-164: a backup nobody has restored is a claim, not a capability

clevervi:
Closes #122

The acceptance criteria are duplicated here on purpose, mapped to evidence, because a criterion that only exists in an issue view is a criterion somebody has to go and find.

## Criterion → evidence

| Acceptance criterion | Evidence | Command |
| --- | --- | --- |
| A documented way to take a backup and to restore one | `docs/operations/backup-and-restore.md` | — |
| A restore actually performed into a temporary database | 31/31 drill checks against the live database | `pnpm db:restore-drill` |
| Verified by content, not exit code: row counts, a sample ID with its history | 4,280 products / 7,639 events / 534 organizations matched; `TP1-2P7M…` resolved with its party, verification state and event count | drill checks 10–24 |
| Written where the other database procedures live | `docs/operations/` | — |
| Roles come back, and the three exist with the right attributes | `roles.sql` asserted to name all three; role attributes compared source↔restored | drill checks 5–7, 27 |
| **`least-privilege.integration.test.ts` passes against the restored database** | **53 tests, connecting as `trustpass_runtime` to `trustpass_restore_drill`** | see below |
| `assertConnectionIsUnprivileged` passes against the restored database | included in those 53 (`connection-privileges.integration.test.ts`) | see below |
| A catalogue comparison, not a spot check | tables, functions, **enabled** triggers, constraints, indexes, every enum value | drill checks 9–14 |
| The temporary database is created empty and never migrated | asserted empty immediately before the restore; `migration_executed: false` in the report; `db:migrate` appears nowhere in the drill | drill check 8 |
| Enough data to represent the model | 476 grants, 170 revocations, 850 memberships, 782 actors, 170 credentials — all compared | drill checks 17–24 |

```
$ pnpm db:restore-drill
31/31 checks passed
Evidence: packages/db/.backups/drill/restore-report.json

$ RUNTIME_DATABASE_URL=…@localhost:5433/trustpass_restore_drill \
  pnpm --filter @trustpass/db test -- src/schema/least-privilege src/connection-privileges
Test Files  3 passed (3)
Tests      53 passed (53)
```

## The finding that justifies the whole issue

`pg_dump` does not write everything TrustPass depends on. One gap was expected. **The other passed 28 of 28 checks and was caught only by running the real suite against the real restore.**

```
source:   PUBLIC connect=false temp=false
restored: PUBLIC connect=true  temp=true
```

`createdb` produces the Postgres default — PUBLIC holding `CONNECT` and `TEMPORARY` — and migration `0024` had revoked exactly that. Every product came back. The boundary #119 and #127 built did not. `least-privilege` said so by failing on `cannot create anything in the schema`.

**That is the issue's thesis, demonstrated rather than argued**: a restore that returns the data and loses the posture is not a recovery.

So the backup is three files:

| File | Carries | Why `pg_dump` does not |
| --- | --- | --- |
| `trustpass.dump` | schema, data, ownership, table grants | — |
| `roles.sql` | the three roles, attributes, memberships | cluster-level, not database objects |
| `database-acl.sql` | the database's own privileges | needs `--create`, which also fixes the database name |

The third is reconstructed from `aclexplode` rather than dumped, because restoring under a different name is the normal case for a drill.

## Two design decisions worth challenging

**The target is created empty and never migrated**, and that assertion is load-bearing: everything present afterwards came out of the backup because nothing else ran. A drill that migrates proves the migrations work — already known, and not the question.

**`--no-owner` is not used.** It makes a restore quieter by discarding ownership, and ownership is what ADR 0013 rests on.

## Two bugs found while building it

**`pg_restore` reading a custom-format dump from a pipe** fails with `did not find magic string in file header`. The format is read by seeking and `/dev/stdin` cannot seek — it reads like a corrupt backup and is not one. Measured, then fixed by copying the file in, which is what a real recovery does anyway.

**The ACL comparison used `ORDER BY 1` inside `string_agg`**, which orders by an argument expression rather than the result. Two identical ACLs compared unequal. Order-dependent, not wrong — **the third time this repository has met that distinction**, after #78 and #131.

## Stated limitations

- **A different cluster is not rehearsed.** Everything restores into the same Postgres, so `roles.sql` is verified by content rather than applied to a cluster that lacks the roles. Restoring onto fresh hardware is described and untested.
- **No RTO or RPO is claimed.** The report records start and finish times; a number measured once on a laptop is not a recovery objective.
- **The drill is not in CI.** It needs `docker exec` against the server's own container to get a client matching the server, and wiring that to a GitHub Actions service container is its own piece of work. Run before a release, and before anything that changes roles or grants.

## Verification

```
pnpm lint / typecheck / build        clean
pnpm test                            538 passed
node --test .claude/tests/*.test.mjs  57 passed
```

## Review bar


clevervi:
Self-review, and what changed after the first push.

## The scanner was right, and it named this exact situation

SonarCloud rated the new code **D** on five findings of *"Path Traversal via faulty LLM-supplied CLI arguments"*. That is not a generic framing here — these scripts are run by an agent assembling a command line, and a positional argument was enough to decide that a dump of the whole database belonged outside the repository.

```
$ pnpm db:backup ../../../../tmp/evil
Refusing to write a backup to C:\tmp\evil, which is outside C:\Users\darkt\Desktop\Project Atlas.
Set TP_BACKUP_ROOT to write somewhere else on purpose.
```

An operator writing to a mounted volume sets that variable once, deliberately, instead of it being a consequence of how a command happened to be built.

It also caught a real race at CRITICAL: the dump went to a fixed path under `/tmp` inside the container, which is world-writable, so anything else there could swap the file between the copy and the restore — the same shape CodeQL found in the preflight. A fresh directory from `mktemp -d` removes the window rather than narrowing it.

Four MINOR findings on PATH lookup for `docker` were fixed too rather than argued down. PATH is a list of directories something else may be able to write to, and these scripts hand a superuser connection to whatever they spawn.

**13 findings → 0.** Gate green.

## And a bug in the fix, of a kind this repository keeps meeting

The absolute-path list did not work. Written with backslashes, it arrived as:

```
C:Program FilesDockerDockeresourcesindocker.exe
```

A backslash followed by `P`, `r` or `b` is an escape sequence, and the rest vanished. It surfaced as *"docker not found at any standard location"* — a true sentence about a path that was never the path. **The fourth time a backslash has been eaten between a writer and a reader here.**

The comment written to explain it was mangled by the same layer on the way in. That is better evidence than the explanation: the answer is to stop writing them. Forward slashes, which Node accepts on Windows and which nothing eats.

## What I would still challenge

**The drill is not in CI**, so it runs when somebody remembers — which is the property this repository keeps replacing with a check. It needs `docker exec` against the server container to get a client matching the server, and wiring that to a GitHub Actions service container is its own piece of work rather than a rider on this. Until then it is a release-time and grant-change-time step, and the `restore-report.json` is what makes a claim about it checkable.

**`roles.sql` is verified by content, not by being applied.** Everything restores into the same cluster, where the roles already exist, so the file is asserted to name all three rather than proven to recreate them. Restoring onto fresh hardware is the case this procedure describes and has not rehearsed, and that is stated in the document rather than left to be discovered.

**No RTO or RPO is claimed.** The report records start and finish times. A number measured once on a laptop is not a recovery objective, and writing one down would be the kind of claim this issue exists to refuse.

Issue #141, opened 2026-09-18
[SECURITY] TP-176: nothing authenticates, and TP-141 already resolves to something else


clevervi:
Found while reading #120 and its dependencies before writing any code, which is the only reason it was found.

## Two problems, one identifier

`#120` says:

> **Depends on** `TP-141`. "Entitled to the answer" needs an actor, and `actor_kind` is still self-declared.

`TP-141` resolves to **#102, closed** — *"ADR — who may speak, and why that authority was still valid when they spoke"*. That ADR shipped in v0.5.0.

But `docs/ROADMAP.md` uses the same identifier for something else entirely:

> **Not shipped: authentication.** `actor_kind` is still self-declared. The model now says exactly what `TP-141` has to build, which it did not before.

So one identifier names two pieces of work: an ADR that is done, and an implementation that was never filed. A reader following `#120`'s dependency finds a closed issue and concludes `#120` is unblocked.

**It is not.** Measured:

```
rg -i "bearer|jwt|authorization header|apiKey|verifyToken|authenticate" apps/api/src
-> nothing
```

There is no authentication of any kind. `POST /products` accepts a registration against any organization, and `POST /enrolments` accepts `actor_kind` as whatever the caller says it is.

`CONTRIBUTING.md` has a rule that a `TP-` identifier must resolve to an issue or a roadmap entry. It does not cover an identifier resolving to **two different pieces of work**, which is worse than resolving to none: absence is visible, and ambiguity reads as completion.

## Why this is `risk:critical` and not merely p1

Every guarantee built in #119, #126, #128 and #130 protects the record from a compromised *runtime*. None of them decides whether the caller is who they say they are. `actor_kind` is self-declared, so:

- a caller can enrol as `holder` for a product they have never held;
- a caller can register against an organization they have no relationship with;
- ADR 0011 pins authority to the moment it was used, and the authority it pins is a claim the caller made about itself.

The roadmap already says the API must not be exposed publicly before this exists. This issue is that sentence given somewhere to live.

## Scope

Authentication of the caller, and nothing else.

- [ ] A caller presents a credential, and the API establishes which `actor` it is.
- [ ] `actor_kind` stops being self-declared. It comes from the authenticated actor, and a request that asserts one is refused rather than believed.
- [ ] `credential` — the table that has existed since v0.5.0 and is read-only to the runtime — is what the credential is checked against.
[...]

PR #150, opened 2026-09-19
[SECURITY] TP-176: a credential says who, and a write now needs one

clevervi:
Implements the authentication core for #141. Refs #141 — it does **not** close it: provisioning and authorization are both still open, and both are listed below and on the issue.

Deliberately not `Closes`. A partial close that GitHub might act on is worse than no keyword at all, and the issue stays open until authentication is provisioned end to end.

Before this branch, `rg -i "bearer|jwt|authorization header|apiKey|verifyToken|authenticate" apps/api/src` returned nothing. Either write endpoint accepted a request with no authenticated caller — the payload validation was there and did its job; what was missing was any answer to *who is asking*.

## What changed

```
tp.<env>.<handle>.<secret>
   |     |        |
   |     |        └─ 256 bits from crypto.randomBytes, SHA-256 at rest, in Node
   |     └────────── 64 bits, stored in clear, UNIQUE, the indexed lookup
   └──────────────── environment, checked before any query runs
```

- `credential` gains `handle` and `secret_digest` — the hole the table has carried on purpose since v0.5.0.
- A verifier resolves a presented token to `{ actorId, credentialId }` or to nothing.
- A middleware refuses any write without one, after CORS and not before.
- ADR 0014 is the design record.

## Three findings worth reading

**The ADR's own token format could not parse itself.** It specified `tp_<env>_<handle>_<secret>`, and `_` is in the base64url alphabet, so a secret containing one split into five parts and was refused by the parser that produced it. Measured before changing anything: **481 of 1000** random 32-byte secrets contain an underscore.

It presented as an intermittent test failure, because half the tokens worked. The rule that outlives the fix is in the ADR: *a delimiter may not be a character the payload alphabet can produce.* The test now issues five hundred and asserts the sample contained the dangerous character, because a loop that gets lucky proves nothing.

**Postgres refused the first constraint against 180 rows.** `(kind = 'api_key') = (handle IS NOT NULL)` is symmetric, and credentials created while this table stored no secret cannot satisfy it. Satisfying it meant inventing a digest or writing `revoked_at` on rows nobody revoked — and per ADR 0011 §7 a revocation is a fact that it happened, not a way to mark a row unusable. The constraint now runs one way. Those rows stay legal and stay unreachable, because verification looks a credential up **by handle** and a row without one is off every path that authenticates.

**An end-to-end test proved less than it claimed.** Three things stop a body's identity reaching an enrolment row and each is sufficient alone, so removing any one left every integration test green; only removing all three turned two red. The single-layer tests now live where the layer does, and the service's field-by-field construction is recorded as redundancy rather than counted as a guard.

## Executed, not asserted

Against a running server and a real database:

```
valid credential                      201
one character changed in the secret   401
the same token presented as live      401
revoked in the database, no restart   401
```

The last is the property ADR 0014 §4 rejected JWT to preserve: `revoked_at` is consulted per request, not cached at startup.

Eight constraint probes against Postgres 18.6, each refusing by its own name:

```
valid api_key, handle and 32-byte digest       ACCEPTED
the same handle twice                          23505 credential_handle_key
a public_key carrying a bearer handle          23514 credential_secret_belongs_to_api_key
a handle with no digest                        23514 credential_secret_is_whole
a digest with no handle                        23514 credential_secret_is_whole
a 31-byte digest                               23514 credential_digest_is_sha256
a 10-character handle                          23514 credential_handle_is_whole
an api_key with neither, as history has them   ACCEPTED
```

The runtime grant was checked rather than assumed: `GRANT SELECT ON TABLE credential` is table-level, so it covers columns added afterwards. As `trustpass_runtime`, reading both new columns succeeds and INSERT is still 42501.

## Mutations

| Mutation | Named test that went red |
| --- | --- |
| delimiter back to an underscore | parses back every token it issues, not most of them |
| part count loosened to at-least-four | does not truncate a secret at a delimiter |
| length checks dropped | refuses a short handle / a short secret |
| environment binding dropped | binds to one environment |
| alphabet check dropped | refuses a handle with characters base64url does not use |
| digest length guard dropped | returns false rather than throwing on a length mismatch |
| revocation no longer checked | refuses a revoked credential |
| expiry `>=` instead of `>` | treats a credential expiring exactly at now() as expired |
| handle ignored | resolves each credential to its own actor and not to any other |
| environment checked after the query | issues no query for anything it can refuse without one |
| a null expiry read as expired | resolves an active credential to its own actor |
| the digest never compared | refuses the right handle with the wrong secret |
| /products not behind a credential | refuses /products with … (9 shapes) |
| /enrolments not behind a credential | refuses /enrolments with … (9 shapes) |
| header split loosened to at-least-two | yields nothing for a third segment |
| the whole header lower-cased | accepts the scheme in any case |
| the principal never set | reaches the handler |
| credential registered before cors | answers a preflight without one |
| principal falls back to the request body | refuses rather than falling back when there is no credential |
| schema stops stripping unknown keys | hands the service four fields, whatever the body contained |
| repository honours what it is handed | ignores an issuer, an origin and a status handed to it anyway |

Three mutants did not compile on the first attempt and the suite failed for the wrong reason. A mutant that breaks the build measures nothing, and reading it as a kill is the same mistake as reading a skipped test as a pass.

One mutant has no entry and the absence is deliberate: *middleware verifies, handler ignores the principal*. No handler reads it yet, so there is nothing to ignore. That test arrives with the first handler that uses it.

## What this does not do

- **Provisioning.** The operator command and its TTY and `CI` barriers are a separate pull request with a different operational risk.
- **Authorization.** `POST /products` still takes `issuer` from its body and records that organization, so an authenticated caller may name one it has no relationship with. Closing it needs actor → membership → organization → grant, which is ADR 0009 and ADR 0011. It is documented on the route, in the OpenAPI description, and asserted by a test that fails the day it changes.
- **Entitlement.** Being the same actor does not prove you created an enrolment. #120.

```
Authentication   who are you              done here
Authorization    what may you do          not started
Entitlement      is this resource yours   #120
```

## Deployment

`TRUSTPASS_ENV` is required with no default. A default is what a forgotten variable falls back to, and a deploy that omitted it would accept tokens minted for staging — the one mistake the environment prefix exists to prevent. It belongs in `.env` and `.env.example`.

## Review

Self-reviewed under the bar in `CONTRIBUTING.md`, which is weaker than independent review: there is one maintainer, `.github/CODEOWNERS` records it, and the mutation table above is what stands in for a second reader.


packages/db/src/auth/verify-credential.ts
```diff
@@ -0,0 +1,100 @@
...
+ * would make "unknown handle" cheaper than "wrong secret" by one SHA-256
+ * comparison. Comparing against this instead keeps the work the same shape.
+ *
+ * It equalises that one step and is not a claim that the whole path is
+ * constant-time — the database round trip dominates, and a lookup that misses
+ * an index is a far larger signal than a digest comparison. What ADR 0014 §7
+ * requires is that the *answer* is identical, and that is enforced by the
+ * single return below.
+ */
+const NEVER_MATCHES = randomBytes(32);
```
clevervi, line 43: **Challenge this one first.** A random value regenerated per process, used so that an unknown handle costs the same SHA-256 comparison as a wrong secret.

What it buys is one step of a path whose cost is dominated by a database round trip, and the honest framing is in the comment: it equalises that step and is not a claim about the whole path. An index miss on `credential_handle_key` would be a far larger signal than a digest comparison.

The alternative was returning early when no row is found — simpler, one less constant, and it makes the two cases measurably different. I took the constant. If you think that is ceremony buying nothing real, the argument is available and the deletion is three lines.

packages/db/src/auth/verify-credential.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * The smallest surface this needs, so a transaction is as acceptable as a pool.
+ *
+ * Typing the parameter as `Database` would exclude the one caller that most
+ * needs it: the test that writes `expires_at = now()` and verifies inside the
+ * same transaction, which is the only place where `>` and `>=` are actually
+ * distinguishable.
+ */
+type Reader = Pick<Database, "select">;
```
clevervi, line 28: Structural rather than nominal, and the reason is one caller: the boundary test writes `expires_at = now()` and verifies **inside the same transaction**, because that is the only place `>` and `>=` are distinguishable — outside one, the clock advances between the two statements and the row is expired either way.

Typing this `Database` would have excluded that caller and the boundary test would have quietly become a tautology. The cost is that this function accepts anything with a `select`, which is looser than the rest of the repository's repository functions, all of which take `Database`.

apps/api/src/env.ts
```diff
@@ -1,10 +1,22 @@
...
+  /**
+   * Which environment a credential must have been minted for — ADR 0014 §3.
+   *
+   * Required, with no default, for the reason the field below spells out at
+   * length: a default is what a forgotten variable falls back to, and the
+   * forgetting is the thing worth catching. A production deploy that omitted
+   * this would otherwise accept tokens minted for staging, which is the one
+   * mistake the prefix exists to prevent.
+   */
+  TRUSTPASS_ENV: z.enum(ENVIRONMENTS),
```
clevervi, line 18: **This breaks a deploy that does not set it, and that is the intent.**

The reasoning is the one already written six lines down for `TRUSTPASS_ALLOW_PRIVILEGED_DATABASE`: a default is what a forgotten variable falls back to, and the forgetting is the thing worth catching. `.default("dev")` would mean a production deploy that omitted this accepts tokens minted for staging — the single mistake the environment prefix exists to prevent.

It needs to land in `.env` and `.env.example` before anyone pulls this. It is configuration discovered while implementing, not a separate defect, so there is no issue for it.

apps/api/src/products/register-product.integration.test.ts
```diff
@@ -82,6 +87,76 @@ describe.skipIf(!databaseUrl)("POST /products against a real database", () => {
     expect(stored?.status).toBe("registered");
   });
 
+  it("records the issuer the body named, which is the boundary #141 does not cross", async () => {
```
clevervi, line 108: This test asserts that a caller can register a product under an organization it has no relationship with, which reads like enshrining a defect.

It is the opposite, and the distinction matters: the gap is real, it is out of scope for authentication, and a limitation nobody wrote down is a limitation nobody fixes. When `actor -> membership -> organization` lands, this test fails, and the failure is the notification.

The alternative was to leave it untested and rely on the prose in `routes/products.ts`. Prose does not fail a build.

clevervi (commented):
Self-review. There is one maintainer and `.github/CODEOWNERS` records it, so this is not a second reader and does not pretend to be — the mutation table in the description is what stands in for one. Four things below are decisions rather than defects, and each is somewhere a different answer was available.

clevervi:
## The five questions about `actor_id`, answered by measurement

### 1. How the existing events stand

```
total_events | actor_id_null | actor_id_set | grant_id_null
       13420 |         13171 |          249 |         13420
```

The 249 that carry one were written after `0026` by the tests on this branch. Every event that predates authentication has `actor_id IS NULL`.

### 2. Nullable

Yes, from `information_schema`:

```
column_name | is_nullable | data_type | column_default
actor_id    | YES         | bigint    |
```

Same pattern as `grant_id`, and for the same reason: a null means *recorded before the system knew who was asking*, which is a true statement about those rows.

### 3. Backfill

None. The migration is three statements:

```sql
ALTER TABLE "lifecycle_event" ADD COLUMN "actor_id" bigint;
ALTER TABLE "lifecycle_event" ADD CONSTRAINT ... FOREIGN KEY ... ON DELETE restrict ON UPDATE cascade;
CREATE INDEX "lifecycle_event_actor_idx" ON "lifecycle_event" USING btree ("actor_id");
```

No `DEFAULT`, no `UPDATE`. The only occurrence of the word in the file is `ON UPDATE cascade` inside the foreign key clause — which turns out to matter, see 5.

Option **A** of the three the review named. Not B: there is no evidence from which a historical actor could be derived, so deriving one would be inventing it.

### 4. Which layer writes it

The repository, from a parameter the service passes from the principal the middleware resolved.

```
credential -> middleware -> principal -> service -> repository parameter -> column
```

Not a trigger, and the boundary is now written in the schema rather than left to be discovered. The reasoning, since the review rightly points at the `recorded_at` precedent:

`recorded_at` and `recorded_in_xact` are server-controlled because **a writer must not be able to choose when it claims to have written** — the database knows the answer independently. There is no equivalent here. The credential is verified in Node and the connection carries no identity, so a trigger could only read a value the application had already told it, via a session variable. A session variable the runtime sets is not a stronger guarantee than a parameter the runtime passes; it is the same trust with more moving parts.

What would change that is a database role per actor. That is not this system, and the comment says where to revisit the decision if it becomes one.

### 5. It cannot be revised — and there was a second door

Four tests, and the second door is the interesting part.

The append-only trigger does cover the new column. Measured by dropping `lifecycle_event_no_update` and watching the two new tests go red alongside the three that were already there:

```
change who an event names               TP002
attribute an event that named nobody    TP002
```

The second is the one that matters for 13,171 rows: nobody can retroactively attribute a historical event.

**But the foreign key is `ON UPDATE CASCADE`.** An actor whose id changed would drag every event naming it along — rewriting attribution without touching `lifecycle_event` at all, and without the trigger having anything to fire on. That door is closed by a different mechanism, so it is asserted separately:

```
renumber an actor that history names    428C9   identity column
delete an actor that history names      23001   restrict
```

Both assertions were `rejects.toThrow()` first, which passes for any error at all. They name the SQLSTATE now, and `428C9` is in `SqlState` with the reason it matters beside it.

---

## Two corrections to my own words

**`actor_kind` is not "what the operation is".** That is `type`. `actor_kind` is the capacity an actor acts under, which is why it lines up with `grant_id` and `organization_id` rather than with `type`. The review is right that the phrasing invites exactly the wrong reading, and it is corrected in the schema and the migration prose with the mistake named rather than quietly replaced.

**A test was missing that no forgery test substitutes for.** Every identity test here showed the body being ignored *when it lies*. None showed that it is not *needed*: a request whose four permitted fields say nothing about who is asking is still attributed. A system that attributed correctly only when the body happened to carry something would have passed every other test in the file.

## On point 10

`actor_id` does not replace `organization_id` and must not start deriving it. They are separate columns with separate constraints, an actor is not an organization, and the relationship between them is exactly what #152 exists to decide. The `/products` test asserts both halves separately today: the actor comes from the credential, the organization comes from the body.

clevervi:
## Correction: the cascade is held by two guards, not one leaning on another

My earlier comment on this pull request said the `ON UPDATE CASCADE` door was closed by `actor.id` being `GENERATED ALWAYS AS IDENTITY`, and that a migration relaxing it "would open it silently". The review reached the same conclusion from the same reasoning, and we were both wrong.

Measured rather than reasoned, this time — relax the identity column, then renumber:

```
identity relaxed to GENERATED BY DEFAULT
renumbering -> TP002  A lifecycle event cannot be modified.
```

**A cascaded UPDATE is an UPDATE**, so the append-only trigger fires on it. The two guards are independent:

```
actor.id GENERATED ALWAYS   ->  428C9, the update never happens
if that is ever relaxed     ->  TP002, the cascade is refused at the history
```

So `ON UPDATE RESTRICT` would be a third layer rather than a repair. **I am not changing it in this pull request**, and the reason is the measurement rather than scope discipline: changing a foreign key to fix a hole that two tests now show is not there would be acting on a theory right after disproving it. If the preference is for the constraint to *state* the intent regardless, that is a defensible schema change and a good follow-up — it just is not a correctness one.

Both guards now have a test, and the second one is new.

## The test that proved it was itself unsafe

Worth reporting in full, because the failure mode is general.

The first version of the new test relied on the failing `UPDATE` to abort the transaction, which rolls back the `ALTER TABLE` with it. That works **exactly while the test passes**.

Running it with `lifecycle_event_no_update` dropped — to check the test was not decoration — the UPDATE succeeded, so the transaction committed, and `actor.id` was left `GENERATED BY DEFAULT` for every run afterwards. One fixture actor was renumbered and three events followed it through the cascade.

A test that weakens the database when its subject is broken is worse than no test. The rollback is unconditional now: the outcome is captured and a sentinel is always thrown, so the DDL cannot commit whatever the database decides. The same mutation re-run confirms it — the test goes red, `actor.id` stays `YES:ALWAYS`.

Then a second, smaller one: the test targeted the literal id `7777777`, and the row the first run had left behind already held it. The `UPDATE` failed with `23505` before reaching the cascade, so the test asserted the wrong refusal for a while. The target is derived now. A memorable constant in a test is a collision waiting for the second run.

## Point 10, confirmed at the row

The property is asserted against Postgres, not against a route:

```
POST /enrolments
  + credential A
  + a body whose four permitted fields say nothing about identity
  -> 201
  -> lifecycle_event.actor_id = A
```

`apps/api/src/enrolments/enrol-product.integration.test.ts`, *"attributes the record when the body says nothing about identity at all"* — it reads the event back through the schema and asserts the id.

## Point 6, agreed and stated as a phase decision

`actor_id` is nullable, so the database alone still permits an authenticated write that names nobody. Today that is prevented in code — the repository parameter is required and the compiler enforces it at twenty-five call sites — and not by a constraint.

That is the right split for this phase and it is a real distinction, so it is recorded as one:

```
0026          legacy compatible, actor_id nullable
this phase    principal required in code, enforced by the type system
later         a database guard, once every legitimate write path can satisfy it
```

The blocker for the third line is that `insertProductWithProvenance` and the status-transition paths also write events, and a `NOT NULL` or a trigger would have to be satisfied by all of them at once.

## Still open

`.env.example` needs `TRUSTPASS_ENV=dev`. I cannot edit it — `.env*` is denied by the repository owner's own tooling permissions, and the denial fired when I tried, which is the guard working. It needs a hand.

clevervi:
## Where the two outstanding items landed

A review of this branch reported that the `ACCEPTED` measurement was not visible and that the branch showed 13 commits. Both were true of the snapshot being read and are not true of the branch now — `c12349c` and `951df5e` landed after it. Pointers, so the next read does not have to hunt:

**The measurement**, `packages/db/src/schema/lifecycle-event.ts` line 214:

```
INSERT product_suspended, actor_kind authority, actor_id NULL  ->  ACCEPTED
```

It replaced the earlier text from `0629546`, which offered two *refused* inserts as evidence. Those refusals came from `lifecycle_event_correction_targets` and TP003 and had nothing to do with `actor_id`, so they proved nothing — which is why the accepted combination was worth finding.

**The test**, `packages/db/src/schema/least-privilege.integration.test.ts` line 335: *"records an event that names nobody, because no constraint asks it to"*. It runs as `trustpass_runtime` against the real role.

It now uses `RETURNING id` and reads that exact row back, asserting `actor_id`, `actor_kind` and `type` on it. The previous version counted rows matching a shape; that count was already scoped to the suspension, because `insertProductWithProvenance` also writes an event with no actor and counting every unattributed row for the product would have passed whether or not the insert under test did anything. Naming the row removes the inference rather than narrowing it.

## Commit count, for the record

15, and it was 14 when a message on this branch claimed 15. Both counts on both sides were wrong at different moments, which is the ordinary hazard of counting a branch that is still moving.

## Still open

`.env.example` needs `TRUSTPASS_ENV=dev`. `.env*` is denied by the repository owner's tooling permissions and the denial fires on every attempt, so it needs a hand. Nothing else on this branch is waiting on anything.

packages/db/src/auth/verify-credential.ts
clevervi: **Kept, and the claim narrowed rather than the code changed.**

Review agreed the equalisation is worth having and, more to the point, agreed that the comment does not oversell it: it equalises one SHA-256 comparison and says so, rather than describing the path as constant-time. The round trip dominates and an index miss would be a far larger signal.

The alternative — returning early when no row is found — stays available and is three lines. Resolving because the decision was examined, not because nobody answered.

packages/db/src/auth/verify-credential.ts
clevervi: **Kept.** The looseness buys the one caller that matters: the boundary test writes `expires_at = now()` and verifies inside the same transaction, which is the only place `>` and `>=` are distinguishable. Typing this `Database` would have excluded it and quietly turned that test into a tautology.

The cost is real and stands: this function accepts anything with a `select`, which is looser than every other repository function here.

apps/api/src/env.ts
clevervi: **Resolved by `c416672`.** `.env.example` now names `TRUSTPASS_ENV=dev`, so a fresh clone can start.

The no-default decision stands and review endorsed it: a default is what a forgotten variable falls back to, and a deploy that omitted this would accept tokens minted for staging — the one mistake the environment prefix exists to prevent.

apps/api/src/products/register-product.integration.test.ts
clevervi: **Kept, and it now has an issue to fail against.** #152 exists, is `priority:p1 / risk:high`, and is assigned to the v0.5.0 milestone. When the actor → membership → organization lookup lands, this test fails, and that failure is the notification.

That was the objection worth raising — a test asserting a gap with no issue behind it reads as approval within months. #153 got the same treatment for the database-level gap.

